PR #45, opened 2015-02-06
Bitvector: Adding support for bit vector literals and operations

ranjitjhala:
See the public API for src/Language/Fixpoint/Bitvector.hs for details.


gridaphobe:
What's the motivation for adding bitvectors? (just curious)

panagosg7:
for typescript. (@ranjitjhala says)

PR #50, opened 2015-02-15
Add type to handle negation of expressions

michaelbeaumont:
Liquid Haskell was having trouble lifting functions with expressions like `-1` to measures. It seemed cleaner to add a real type like `PNot`, called `ENeg`, than to always convert to `0 - 1`.

I'll also make a pull request for the corresponding change in liquidhaskell. 
I can also add the type to the OCaml side of things if this seems like a good idea.

This change requires a rebuild of the `fixpoint` executable.


nikivazou:
Thanks @michaelbeaumont !

PR #73, opened 2015-06-02
Cutsolver

ranjitjhala:
Misc. edits to the parser to fix various things.

@nikivazou see: dd1ae19 -- I am kicking out the special case 
handling of `eNil` and `eCons` from the fixpoint parser. I don't 
want *any* language specific infix operators etc. here, you should:

1. Make up new symbols like: `smt_nil` and `smt_cons` etc.
2. Use SMTLIB style prefix applications e.g. `(smt_cons x xs)` etc.
3. Deal with any syntactic transformations in LiquidHaskell land.

This is analogous to how we let users write `union x y` but 
translate it to `Set_cup x y` in fixpoint.

Specifically  `(x:y)` was causing problems because `x : int` is supposed to be parsed as `Cast x int`
and not `eCons x int`...


ranjitjhala:
@nikivazou recall this is going to break LH `test/{pos, neg}/ListDataCons.hs`

PR #89, opened 2015-07-28
Eliminate.hs : variable names and readability

atondwal:
`foo`, `baz`, `blah`, `go` have been replaced by more meaningful names *`predsBinds`, `substBinds`, `substBind`, `replacement`* respectively.

replaced some lambdas and do-blocks with more idiomatic functions from `base` (in particular `\(x,y) -> (x,f y)` have been replaced by `second f` everywhere)


ranjitjhala:
Thanks!

Issue #101, opened 2015-08-21
Fixpoint parser produces a different result than liquid


christetreault:
The `FInfo` that `fixpoint` produces when reading in a `.fq` file is not the same as the `FInfo` that `liquid` produces when reading in a `.hs` file. This is causing issues for many files in the `liquid` test suite.

Given the following input:

```
fixpoint +RTS -N4 -RTS --cores=4 ../liquidhaskell/tests/pos/.liquid/RBTree-col-height.hs.fq
```

A [massive error](https://gist.github.com/christetreault/b0c480b8efb70ae0e3fd#file-gistfile1-txt) is returned. 

Much discussion of this issue has happened in the thread for issue #90:

https://github.com/ucsd-progsys/liquid-fixpoint/issues/90#issuecomment-131186394 by Chris
 ... through ...
https://github.com/ucsd-progsys/liquid-fixpoint/issues/90#issuecomment-131921379 by Chris

ranjitjhala:
@christetreault can you let me know if this issue still remains on the new `master` branch. (I did a bunch of changes to the parser/serializer that *should* have addressed this.)

christetreault:
@ranjitjhala 

The follwing tests pass for me on the current fixpoint/liquid masters:

neg/pred.hs
neg/sumk.hs
crash/*
error/crash/*

All other tests fail with ExitCode 2, and I'm getting parser errors when running liquid manually

christetreault:
Still no luck, now I get this:

```
$ fixpoint --cores=4 ../liquidhaskell/tests/pos/.liquid/RBTree-col-height.hs.fq
Liquid-Fixpoint Copyright 2009-13 Regents of the University of California.
All Rights Reserved.

Unexpected Error: ../liquidhaskell/tests/pos/.liquid/RBTree-col-height.hs.fq:406:21-406:21: Error: parseError "../liquidhaskell/tests/pos/.liquid/RBTree-col-height.hs.fq" (line 406, column 21):
unexpected "("
expecting ")"
 when parsing from ../liquidhaskell/tests/pos/.liquid/RBTree-col-height.hs.fq

EXIT: ExitFailure 2
```

At least the error is more useful!

bmcfluff:
I think the problem is with type variable application (i.e. "@(0) @(1)") - see https://github.com/ucsd-progsys/liquid-fixpoint/issues/75#issuecomment-135568522 - I think that "step 4" does need to be implemented in the Haskell parser.

[...]

Issue #90, opened 2015-07-31
Implement parallel FInfo solver


christetreault:
Given that we have functions to partition an `FInfo` into a list of `FInfo`s, it makes sense to implement a parallel version of `solveExt`; `solvePar`.

Work will be done in the branch: https://github.com/ucsd-progsys/liquid-fixpoint/tree/cores
any related functionality in liquidhaskell will be done in: https://github.com/ucsd-progsys/liquidhaskell/tree/cores

**NOTE:** I'm opening this issue to have a central location to post updates. This we aren't sending emails randomly throughout the group.

christetreault:
Today, I did some informal testing. I ran all the files in the liquidhaskell test suite through `time`, both in serial and parallel. Results that I found _noteworthy_ can be found at: https://drive.google.com/file/d/0Bx6BpheCXcBWR3dnVS1BYk9uQzg/view?usp=sharing

Some particularly interesting results:

```
./pos/RBTree-color.hs:real 118.07 user 302.39 sys 45.61	      |	./pos/RBTree-color.hs:real 9.57 user 7.35 sys 0.61

slower: 234 finfos

./pos/RBTree-height.hs:real 90.52 user 229.64 sys 35.69	      |	./pos/RBTree-height.hs:real 10.85 user 6.55 sys 0.66

slower: 263 finfos

./pos/RBTree-col-height.hs:real 239.64 user 616.80 sys 90.52  |	./pos/RBTree-col-height.hs:real 21.78 user 12.22 sys 1.22

slower: 404 finfos
```

These red black tree files all generated 200+ FInfos, with col-height generating 404. For all of these, the parallel solver was unusably slow.

```
./pos/BST.hs:real 15.33 user 15.40 sys 1.96		      |	./pos/BST.hs:real 21.40 user 12.44 sys 0.97

faster: 5 finfos

./pos/Map.hs:real 45.55 user 40.94 sys 3.38		      |	./pos/Map.hs:real 63.54 user 39.23 sys 1.89

faster: 2 finfos

./pos/Map0.hs:real 44.21 user 39.10 sys 3.80		      |	./pos/Map0.hs:real 63.86 user 39.40 sys 2.04

faster: 2 finfos

./pos/Map2.hs:real 45.69 user 39.65 sys 3.60		      |	./pos/Map2.hs:real 53.16 user 38.27 sys 1.74

faster: 2 finfos
```

[...]

PR #102, opened 2015-08-24
Merge cores into master

christetreault:
I believe that the Parallel solver is ready to be merged into master. The following items remain:

issue #101: The fixpoint parser has some issue that becomes apparent when reading in and solving a .fq file in parallel. This issue, while related to the parallel solver, is not actually caused by the parallel solver.

Default minPartSize: The default minimum partition size is currently 1. This default represents a performance loss in basically any scenario, however it is this way currently due to the way that liquid handles configurations. This should be opened as an issue on the liquid bug tracker, but I don't believe that we need to block on this.

`cPartSize`: We are determining the "size" of a `CPart` using a very rudimentary metric. It works for the most part, but a performance gain could likely be realized by improving this function. However, this is a [hard problem](https://github.com/ucsd-progsys/liquid-fixpoint/issues/90#issuecomment-126841619), so I don't think we should block on it.

---

I don't believe that any of these issues should prevent the parallel solver from being merged into master.


gridaphobe:
Would you mind updating the PR to build with GHC 7.8.4? (See https://travis-ci.org/ucsd-progsys/liquid-fixpoint/jobs/77023675#L351-L360, it looks like a missing import)

christetreault:
OK @gridaphobe looks like we're good!

PR #123, opened 2015-09-08
factored `defaultFoo` functions out of liquid

christetreault:
Adding `defaultCores`, `defaultMinPartSize`, and `defaultMaxPartSize` to the export list, to be consumed by liquidhaskell.

Previously liquidhaskell had duplicate versions of these, and the two required synchronization.


christetreault:
depends on: https://github.com/ucsd-progsys/liquidhaskell/pull/445

christetreault:
merged into make-par-default: https://github.com/ucsd-progsys/liquid-fixpoint/pull/129

PR #124, opened 2015-09-09
Improve the measure of |partition|

christetreault:
Partition size is now a function of `|WfC list|` as well as the `|SubC map|`. This seems to be realizing a speedup on the order of ~30 seconds for sufficiently large files.


christetreault:
Merged into make-par-default: https://github.com/ucsd-progsys/liquid-fixpoint/pull/128

PR #130, opened 2015-09-10
Set parallel solving be be default

christetreault:
The solver now by default solves on the maximum number of available cores. Single threaded solving can be accomplished by setting `--cores=1`. Any number of cores passed to `--cores=X` will override the default behavior.


christetreault:
The following pull request is dependent on this one: https://github.com/ucsd-progsys/liquidhaskell/pull/450

PR #141, opened 2015-09-21
Remove gs from GInfo

bmcfluff:
Condenses 'gs' and 'lits' fields into a single field containing all fq 'constants'.


bmcfluff:
This PR goes with https://github.com/ucsd-progsys/liquidhaskell/pull/462 (and we should bump version numbers accordingly)

ranjitjhala:
@gridaphobe  and @christetreault this is a BREAKING change for LH. Can you please advise
as to how we should bump up version numbers etc?

christetreault:
If it's breaking, I believe the correct thing to do would be to bump from `0.4.0.0` -> `0.5.0.0`.

I'll put in a PR for this.

christetreault:
...done. See PR https://github.com/ucsd-progsys/liquid-fixpoint/pull/143

PR #143, opened 2015-09-22
Bumping version number to 0.5

christetreault:
OK, break it all you want, the version has been bumped. I'll make LH in the repo depend on `liquid-fixpoint == 0.5.*` so it won't try to install from hackage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,50 @@ git clone git@github.com:ucsd-progsys/liquid-fixpoint.git
 Configuration Management
 ------------------------
 
-It is very important that the version of
+It is very important that the version of Liquid Fixpoint be maintained properly.
+
+Suppose that the current version of Liquid Haskell is `A.B.C.D`:
+
++ After a release to hackage is made, the `D` component of Liquid Fixpoint shall be incremented by 1. The version of Liquid Fixpoint is now `A.B.C.(D + 1)`
+
++ The first time a new function or type is exported from Liquid Fixpoint, the `C` component shall be incremented, and the `D` component shall be set to `0`. The version of Liquid Fixpoint is now `A.B.(C + 1).0`
+
++ The first time the signature of an exported function or type is changed, or an exported function or type is removed, the `B` component shall be incremented by 1, and the `C` and `D` components shall be set to `0`. The version of Liquid Fixpoint is now `A.(B + 1).0.0`
+
++ The `A` component shall be updated at the sole discretion of the project owners.
+
+It is recommended to use the [Bumper](https://hackage.haskell.org/package/bumper) utility to manage the versioning of Liquid Fixpoint. Bumper will automatically do the correct update to the cabal file. Additionally, it will update any packages that you have the source for that depend on Liquid Fixpoint.
+
+To update Liquid Fixpoint and Liquid Haskell, first clone Liquid Haskell and Liquid Fixpoint to a common location:
+
+```
+git clone https://github.com/ucsd-progsys/liquidhaskell.git
+git clone https://github.com/ucsd-progsys/liquid-fixpoint.git
+```
+
+To increment the `D` component of Liquid Fixpoint:
+
+```
+./path/to/bumper -3 liquid-fixpoint
+```
+
+This will update the `D` component of Liquid Fixpoint. If necessary, this will update the `Build-Depends` of Liquid Haskell. If the `Build-Depends` was updated, Liquid Haskell's `D` component will be incremented.
+
+To increment the `C` component of Liquid Fixpoint, and to set its `D` component to `0`:
+
+```
+./path/to/bumper --minor liquid-fixpoint
+```
+
+As before, this will update Liquid Fixpoint and, if necessary, Liquid Haskell.
+
+To increment the `B` component of Liquid Fixpoint, and to set its `D` and `C` components to 0:
+
+```
+./path/to/bumper --major liquid-fixpoint
+```
+
+As before, this will update Liquid Fixpoint and, if necessary, Liquid Haskell
 
 SMTLIB2 Interface
 -----------------
--- a/README.md
+++ b/README.md
@@ -81,11 +81,11 @@ It is very important that the version of Liquid Fixpoint be maintained properly.
 
 Suppose that the current version of Liquid Haskell is `A.B.C.D`:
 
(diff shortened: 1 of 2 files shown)


ranjitjhala:
Super -- just to be sure, you will (have)  also bump(ed) the LH version
right?  Thanks!

In general, can you draft a small para explaining "the protocol" and add it
to the README
for liquid-fixpoint?

Thanks!

Ranjit.


On Tuesday, September 22, 2015, Chris Tetreault <notifications@github.com>
wrote:

> OK, break it all you want, the version has been bumped. I'll make LH in
> the repo depend on liquid-fixpoint == 0.5.* So it won't try to install
> from hackage.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/ucsd-progsys/liquid-fixpoint/pull/143
> Commit Summary
>
>    - Bumping version number to 0.5
>
> File Changes
>
>    - *M* liquid-fixpoint.cabal
>    <https://github.com/ucsd-progsys/liquid-fixpoint/pull/143/files#diff-0>
>    (2)
>
> Patch Links:
>
>    - https://github.com/ucsd-progsys/liquid-fixpoint/pull/143.patch
>    - https://github.com/ucsd-progsys/liquid-fixpoint/pull/143.diff
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ucsd-progsys/liquid-fixpoint/pull/143>.
>


-- 
Ranjit.

gridaphobe:
There's no need to bump the LH major version if we haven't (yet) broken
anything. A minor bump would be in order though.

There's a bit more than a paragraph at http://semver.org (Haskell's PVP
is slightly different, but the basic idea is the same IIRC).

ranjitjhala:
Ah, I'm not looking for a detailed description; just a CHECKLIST of
Instructions of the form:

+ In repo X change file Y to Z
...

christetreault:
I find the PVP versioning to be strange, personally. If I understand correctly, for version `A.B.C.D`, `A.B` is for breaking changes, `C` is for non-breaking additions, and `D` is for "whatever you feel like doing, GOSH!" That said, it's the official policy of "our community" so I imagine we should follow that instead of the usual semantic versioning.

I'm thinking the versioning instructions can say something to the effect of "When a release is made to hackage, bump digit `C` by 1. If a breaking change is made to the repo, set `C` to 0, and bump `B` by 1." `A` will of course be sacred and only modifiable by a grownup.

christetreault:
@ranjitjhala I'll bump the version of liquid haskell as part of my changes for https://github.com/ucsd-progsys/liquidhaskell/issues/461 which will hopefully be done today. I had made progress yesterday, but managed to break my haskell installation.

gridaphobe:
Ooh [bumper](https://github.com/silkapp/bumper/blob/master/README.md) looks super useful.

Here's roughly how it should be used, assuming you're in the "top-level" liquid directory, i.e.

```
liquid
 |- liquid-fixpoint
 |- liquidhaskell
 |- refscript
 \- etc.
```

- If you change any exported type signature: `bumper --major liquid-fixpoint`
- If you add a new exported function: `bumper --minor liquid-fixpoint`
- Upon releasing to hackage: `bumper -3 liquid-fixpoint`

bumper will supposedly handle the transitive version bumps to liquidhaskell and other packages.

christetreault:
Is refscript something I need to be concerned about? How many projects do we control that have a dependency on fixpoint?

gridaphobe:
refscript isn't on hackage, so not really

http://packdeps.haskellers.com/reverse/liquid-fixpoint shows that
liquidhaskell and target are the only released projects. i develop
target against the hackage release of fixpoint/LH (which is much more
sane IMO) so you don't need to worry about it either.

christetreault:
This bumper thing seems reasonable. I'll give it a shot and see how it goes. I can probably write a few bash scripts to do the right thing if it works out and our documentation for versioning will be "run the script"

christetreault:
We don't re-export anything from fixpoint in liquid do we?

ranjitjhala:
Yes we should worry about RefScript -- as in this policy should be applied across the board regardless of whether or not the package is (yet) on hackage.

christetreault:
OK, I don't seem to have access to that repo, so I'm going to need to get added if I'm going to include it.

christetreault:
Bumper seems to work. I'll go about updating the docs to use it. I'm not sure it's worth writing a script to do `./bumper --major liquid-fixpoint` so I'll probably skip that part.

bmcfluff:
@christetreault - by re-exporting, are you referring e.g. to "isDummy" (https://github.com/ucsd-progsys/liquid-fixpoint/blob/master/src/Language/Fixpoint/Types.hs#L165) (https://github.com/ucsd-progsys/liquidhaskell/blob/master/src/Language/Haskell/Liquid/Types.hs#L39)? Not sure why we're doing that (and now might be the time to fix it...)

christetreault:
Yeah, that is what I'm talking about. If the signature of that is changed by fixpoint, then liquid has to get a major version bump.

That said, I think the way to go is to have separate versioning instructions for liquid, similar to what I'm doing right now in fixpoint. Those instructions can address that, since we don't maintain anything that fixpoint depends on, we don't have to worry about these issues.

christetreault:
OK, how does this look? https://github.com/ucsd-progsys/liquid-fixpoint/tree/bump-version#configuration-management

bmcfluff:
The bumper utility doesn't seem to put trailing 0s on version numbers.

christetreault:
Yeah, but that's purely aesthetic. I guess I could update the instructions to match that, but I think people will figure it out.

christetreault:
Hmm... This might actually be a real problem...

https://github.com/haskell/cabal/issues/1368

christetreault:
```
Prelude Data.Version> Version [0,5] [] >= Version [0,5,0] []
False
Prelude Data.Version> Version [0,5] [] <= Version [0,5,0] []
True
Prelude Data.Version> Version [0,5] [] == Version [0,5,0] []
False

```

I'll update the docs to specify stripping the trailing zeros.

christetreault:
OK, thanks @bmcfluff for thinking of that. Now the instructions specify stripping trailing zeros, so we shouldn't run into this issue.

PR #145, opened 2015-09-27
Fix inst

ranjitjhala:
Reduce number of unification calls (to `unifyWith`) needed during qualifier instantiation, by:

1. Grouping binder by `Sort.t`,
2. Using equality on `Sort.t` when the (qualifier parameter) is monomorphic (i.e. has no free type variables.)


ranjitjhala:
I had to do a fair bit of surgery to resolve conflicts.

PR #148, opened 2015-10-02
Changed Subst internals from association list to HashMap

bmcfluff:


ranjitjhala:
Can we undo this, for reasons I find hard to articulate, I think this is very problematic. 

(Copied from slack)

@benjamincosman — there is something bizarre happening, see the `binary` branch. 

I have added a `—binary` flag such that:
`fixpoint —binary foo.fq` simply SAVES the `.fq` file to a `.bfq` file. Now subsequently one can run either `fixpoint —native foo.bfq` (or even `fixpoint foo.bfq` to use the ocaml solver) The OCAML solver works fine on the BFQ file, but the `—native` solver fails with wierd “non unique-sort” errors. I THINK this has to do with the recent change to using MAPS for Substitutions, there was a reason they were implemented with LISTs instead, I think had to do with the fact that you want to preserve ordering or somesuch (i.e. the substitutions TELESCOPE, i.e. `[x := y] [y := z]`  is actually equivalent to the TELESCOPED `[x := z] [ y := z]` BUT  .`[y := z][x := y] ` is NOT equivalent to  the TELESCOPED variant. Perhaps this INVARIANT has not been maintained — or rather the `catSubst` using the `M.union` is guaranteed to jumble the order of the substitutions which causes all sorts of wierd problems. Can you revert that to LISTS and see if this issue gets resolved? Thanks!

gridaphobe:
Can't we just maintain the invariant by applying the new substitution to the old one before we union them?

ranjitjhala:
That should work ...

bmcfluff:
That's exactly what I'm doing, right?

```
unsafeCatSubst (Su s1) θ2@(Su s2) = Su $ M.union s1' s2
  where
    s1'                           = subst θ2 <$> s1
```

ranjitjhala:
Yes looks like it...

gridaphobe:
Yes it is. Also noteworthy is that the list-based `unsafeCatSubst` was **also** applying `s2` to `s1` before combining them.

Question: do we have an invariant that we only combine substitutions whose domains are disjoint? If not, what are the expected semantics of an overlapping `catSubst`? I might expect the "new" `s2` to take precedence, but this code (and the list version) would do the opposite.

nikivazou:
This sounds very familiar, in that I did deal this this issue a long time ago. 
The domains are definitely not always disjoint and, if I remember correctly, union does not combine them with the "expected" way. So, fixpoint expects the "old" should take precedence.

ranjitjhala:
Yes OLD takes precedence not new (if you think about it that's what's sensible) and domains may not be disjoint. Unfortunately this whole thing is rather tricky...

spinda:
I would expect `subst (catSubst su1 su2) expr` to have the same effect as `subst su2 (subst su1 expr)`.

____

Say `su1` is `[y := z]`, `su2` is `[y := q]`, and `expr` is `x + y`.

`subst su1 expr` should yield `x + z`. Then `subst su2 (subst su1 expr)` should yield `x + z` as well, since `y` is  no longer around to be substituted.

Under the current implementation, `catSubst su1 su2` would yield `[y := z]`, as the `y` in `su1` overrides the one in `su2`. So `subst (catSubst su1 su2) expr` would yield the same as the above.

___

Say `su1` is `[x := y]`, `su2` is `[y := z]`, and `expr` is `x + y`.

`subst su1 expr` should yield `y + y`. Then `subst su2 (subst su1 expr)` should yield `z + z`.

Under the current implementation, `catSubst su1 su2` would first apply `su2` to `su1`'s map, yielding `[x := z]`, then union the two maps, yielding `[x := z] [y := z]`. This new substitution would produce the same result as the above if applied to `expr`.

____

But what about when you have something like `su1 = [x := y] [y := z]` in one substitution? If `expr` were `x + y`, should `subst su1 expr` be `y + z` or `z + z`? Both the new and the old implementations would produce `y + z` (but intuitively I would expect `z + z`).

Assume that `subst su1 expr` should indeed be `y + z`. Then, say `su2` is `[y := q]`. `subst su2 (subst su1 expr)` would yield `q + z`.

`catSubst su1 su2` would first apply `su2` to `su1`, yielding `[x := q] [y := z]`. Then the union would produce `[x := q] [y := z]`, and `subst (catSubst su1 su2) expr` would yield `q + z`, as expected.

So the main grey area doesn't seem to be in `catSubst`, but in how `[x := y] [y := z]` should be interpreted.

PR #149, opened 2015-10-02
Deepseq

ranjitjhala:
Remove space leak in `renameVars` -- don't use `foldlM` (and, don't use `State` just for folding).


ranjitjhala:
merging because tests pass locally...

PR #154, opened 2015-10-08
Symbol

ranjitjhala:
Use an `intern` representation for `Symbol`.

### DIFF
--- a/src/Language/Fixpoint/Names.hs
+++ b/src/Language/Fixpoint/Names.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE StandaloneDeriving         #-}
 {-# LANGUAGE TypeSynonymInstances       #-}
 {-# LANGUAGE ViewPatterns               #-}
+{-# LANGUAGE PatternGuards              #-}
 {-# LANGUAGE CPP                        #-}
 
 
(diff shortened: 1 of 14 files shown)


src/Language/Fixpoint/Names.hs
```diff
@@ -20,20 +19,47 @@ module Language.Fixpoint.Names (
...
-  , dropModuleUnique
-  , takeModuleNames
+
+  -- * Destructors
+  -- , stripParensSym
+  -- , stripParens
+  , stripPrefix
+  , consSym
+  , unconsSym
+  , dropSym
```
gridaphobe: I plan to remove most of these functions on `Symbol`, basically anything of type `a -> Symbol` other than `symbol` itself. Constructing a `Symbol` is expensive and we should discourage manipulating `Symbol`s (as opposed to manipulating a `Text` and converting it to a `Symbol` at the end).

src/Language/Fixpoint/Names.hs
```diff
@@ -58,195 +89,261 @@ import           Control.Applicative ((<$>))
...
+
+-- | Invariant: a `SafeText` is made up of:
+--
+--     ['0'..'9'] ++ ['a'...'z'] ++ ['A'..'Z'] ++ '$'
+--
+--   If the original text has ANY other chars, it is represented as:
+--
+--     lq$i
+--
+--   where i is a unique integer (for each text)
```
gridaphobe: I **really** don't like this encoding scheme, I fear it will make the fq files (or god forbid the smt files) WAY harder to read.

I strongly suggest adopting a variant of GHC's [z-encoding](https://ghc.haskell.org/trac/ghc/wiki/Commentary/Compiler/SymbolNames#AlphanumericCharacters), so we can actually read the intermediate files.

PR #170, opened 2015-12-11
Splitting `Types` into a bunch of smaller modules

ranjitjhala:
and generally cleaning up. Much remains todo, but at least those changes should be isolated _within_ the individual modules.


ranjitjhala:
@bmcfluff please pull ASAP, else we'll be in crazy conflict hell...

PR #171, opened 2015-12-11
Remove WrappedC

bmcfluff:


ranjitjhala:
@bmcfluff are all the constraints in the `cm0` in the signature [here](https://github.com/ucsd-progsys/liquid-fixpoint/pull/171/files#diff-08563e0ad9a9edb5cdd5502f320682b8R28) contained inside the `SInfo` ?

ranjitjhala:
Put another way, why do we have a *separate* `cm0` and `SInfo` ? The latter should be "the constraints"...?

bmcfluff:
cm0 is the constraints as passed in from the client. SInfo contains the constraints after we've done stuff to them including renaming variables and pushing the lhs into the environment, changing the type from SubC to SimpC (FInfo to SInfo) in the process. I figured both the client and the type system would prefer to learn which of their original constraints were invalid, not our modified ones. Thoughts?

ranjitjhala:
> I figured both the client and the type system would prefer to learn which
> of their original constraints were invalid, not our modified ones. Thoughts?

Doesn't each _transformed_ constraint map to a unique _input_ constraint 
with the same id? If so, it should suffice to preserve those.

Alternately (note that this is a nice illustration of the utility of parametric polymorphism),
as long as we ensure the top-level solvers in `Solver.hs` preserve the signature:

```
solve :: (NFData a, Fixpoint a) => Solver a
``` 

then the client has all the information they need about which constraint "failed" because
they can index the individual constraints however they please, and the `Result` will tell them
which constraints failed.

bmcfluff:
Ok, I pushed the polymorphic solution.

PR #179, opened 2016-01-13
Relax cereal to >= 0.4; update Stackage snapshot

spinda:
Fixes https://github.com/ucsd-progsys/liquidhaskell/issues/571


ranjitjhala:
Is there a reason to new to use the new resolver? it just triggers a long build...

spinda:
I updated so it would match LiquidHaskell's, as them being out-of-sync was the source of the issue in the first place.

PR #180, opened 2016-01-13
Expose fApp'

spinda:
This is a useful function that I'd like to use from LiquidHaskell, but it's not currently exported.


ranjitjhala:
@nikivazou is playing around with this code a lot may make sense to try to pull from her branch?

spinda:
Is that https://github.com/ucsd-progsys/liquid-fixpoint/tree/genEApp?

PR #183, opened 2016-01-20
implementation of delta debugging minimization

rolph-recto:
* The delta debugging code lives in `Solver.hs`; I can't move it to its own module because it has a circular / mutual dependency with `solve`.

* Use case is `fixpoint --minimize [file]`. The minimization code then outputs `.liquid/[file].minfq`, which is an FQ file that contains the minimal unsatisfiable subtyping constraint set.

* Added cleanup code and exception handling to `runSolverM` in `Solver/Monad.hs`. Multiple calls to the solver, such as what delta debugging minimization does, broke previously because the *.smt2 file was not released.

### DIFF
--- a/src/Language/Fixpoint/Smt/Interface.hs
+++ b/src/Language/Fixpoint/Smt/Interface.hs
@@ -248,9 +248,8 @@ makeProcess s
 --------------------------------------------------------------------------
 cleanupContext :: Context -> IO ExitCode
 --------------------------------------------------------------------------
-cleanupContext me@(Ctx {..})
-  = do smtWrite me "(exit)"
-       code <- waitForProcess pId
+cleanupContext (Ctx {..})
+  = do code <- waitForProcess pId
        hClose cIn
        hClose cOut
        maybe (return ()) hClose cLog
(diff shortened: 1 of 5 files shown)


gridaphobe:
Sorry, one more thing. We need to move the `smtWrite "(exit)"` call out of `cleanupContext`, and into `runSolverM`. Otherwise if z3 dies unexpectedly, I believe we'll get another exception when we try to write to the pipe.

Would you mind making that change?

https://github.com/ucsd-progsys/liquid-fixpoint/blob/cd1a1afaf6f8d9ebd4d63f925aefa580332b5f41/src/Language/Fixpoint/Smt/Interface.hs#L249

rolph-recto:
Done! Just had to add an `OverloadedStrings` pragma in `Monad.hs` to make it compile.

gridaphobe:
thanks!

ranjitjhala:
Is there any documentation/examples on how to run this?

rolph-recto:
Nothing beyond the blurb on the original post (https://github.com/ucsd-progsys/liquid-fixpoint/pull/183#issue-127768997) and the `--help` command. Is that sufficient?

ranjitjhala:
Are there any example tests you have run it on, e.g. a file `tests/min/foo.fq` which when run through the `--minimize` yields `tests/min/foo.minfq` (which is also unsat?)

ranjitjhala:
i.e. can you add one or two such tests?

Issue #193, opened 2016-02-04
Bug in SMT embedding of polymorphic applications


ranjitjhala:
@nikivazou there is some issue in your defunctionalization, see:

https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/tests/todo/poly.fq

which is just:

```
// MONOMORPHIC This definition works fine ...
// constant offset : (func(2, [int ; (BitVec Size32) ]))

// POLYMORPHIC But this crashes as 'offset 0' is embedded as int not bv...
constant offset : (func(2, [int; @(1)]))

bind 0 x  : {VV : (BitVec  Size32) | [ VV = offset 0 ]}

constraint: 
  env [0]
  lhs {VV : (BitVec Size32) | [ VV = x ] }
  rhs {VV : (BitVec Size32) | [ VV = x ] }
  id 1 tag [1]
```

The *monomorphic* definition works fine, but the *polymorphic* one doesn't, because the `offset 0` 
term gets embedded/serialized as `int` using the `int_apply`. This is a bug because the term is infact a `BitVec...` as thats the only way that the equality could be well-typed.

nikivazou:
Pushed a fix!

ranjitjhala:
is the `elaborate` code fixed and merged in? if so, shall we close this issue?

nikivazou:
Yes, it is fixed, closing!

PR #194, opened 2016-02-04
Elaborate Eq

nikivazou:
fix for #193


ranjitjhala:
@nikivazou see the changes. the particular tests here (and the original issue passes) but I need to build against LH etc. to see that nothing is broken...

nikivazou:
Looks good, other than it breaks the prover, but will fix it later. 

I created a liquidhaskell pull request to test lh tests

https://github.com/ucsd-progsys/liquidhaskell/pull/607

PR #196, opened 2016-02-09
Nonlinear

nikivazou:
Make interpreted mul and div the default behavior

### DIFF
--- a/src/Language/Fixpoint/Smt/Interface.hs
+++ b/src/Language/Fixpoint/Smt/Interface.hs
@@ -100,7 +100,8 @@ runCommands cmds
 -- TODO take makeContext's Bool from caller instead of always using False?
 makeZ3Context :: FilePath -> [(Symbol, Sort)] -> IO Context
 makeZ3Context f xts
-  = do me <- makeContext False Z3 f
+  = do me <- makeContextWithSEnv False Z3 f $ fromListSEnv xts
+       smtDecls me (toListSEnv initSMTEnv)
        smtDecls me xts
        return me
 
--- a/src/Language/Fixpoint/Solver/Monad.hs
+++ b/src/Language/Fixpoint/Solver/Monad.hs
@@ -89,12 +89,14 @@ runSolverM cfg fi _ act = do
     return $ fst res
 
   where
-    acquire = makeContextWithSEnv (linear cfg) (solver cfg) file env 
+    acquire = makeContextWithSEnv lar (solver cfg) file env 
     release = cleanupContext
     be      = F.bs     fi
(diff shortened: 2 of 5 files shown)


gridaphobe:
How does this affect SMT solvers other than Z3? IIRC we claim to support CVC4 and MathSat as well.

nikivazou:
Good point! We do not have these solvers in the test suite right? I am adding a `linear` flag so we can switch again in case of a problem

gridaphobe:
I think we only test with Z3 at the moment. My point was more that we should try to ensure all solver backends work well out-of-the-box, e.g. if MathSat doesn't support non-linear arithmetic, then `--smtsolver mathsat` should imply `--linear`.

gridaphobe:
(And this should be documented in the README and CHANGELOG!)

nikivazou:
I bet most smt solvers support mul and div. Though, given that we are going into an (in theory) undecidable logic not sure how efficient they will be.

gridaphobe:
But we shouldn't bet about the solvers we list in the README as supported, we should test! And even if they support interpreted mul and div, but produce sub-optimal results (e.g. timing out or returning "unknown") we should perhaps use the uninterpreted versions.

nikivazou:
Ok, lets see what Z3 does first

ranjitjhala:
@nikivazou -- can you add CSV files (performance) for before / after this
merge?

nikivazou:
I have this https://github.com/ucsd-progsys/liquidhaskell/blob/bab3b940891809a6d7a5efe871faafebd2f784ad/tests/logs/defunctionalized.csv

and the one where the numbers are actually compared:

https://github.com/ucsd-progsys/liquidhaskell/blob/bab3b940891809a6d7a5efe871faafebd2f784ad/tests/logs/slowdown.csv

gridaphobe:
@nikivazou Please update the CHANGELOG (and README if it mentioned `--real`)

nikivazou:
I did in liquidHaskell

On Tue, Feb 9, 2016 at 1:56 PM, Eric Seidel <notifications@github.com>
wrote:

> @nikivazou <https://github.com/nikivazou> Please update the CHANGELOG
> (and README if it mentioned --real)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ucsd-progsys/liquid-fixpoint/pull/196#issuecomment-182094773>
> .
>



-- 
Niki Vazou

gridaphobe:
Fixpoint is a separate package, and has its own changelog :)

nikivazou:
Done!

PR #216, opened 2016-03-23
promote ints to reals instead of dropping refinements

gridaphobe:
Currently we prune refinements like `0 != 0.0`, which is unsound.
We could just report a sort error, but this leads to problems
for liquidhaskell, as `0` can refer to an Int or a Double.
So instead we promote ints to reals when necessary, which is sound.

It feels a bit wrong, but it's much simpler than any alternative
I can think of that doesn't reject ucsd-progsys/liquidhaskell#602
as ill-formed rather than unsafe.

needed to fix ucsd-progsys/liquidhaskell#602

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@
 /tests/neg/.liquid/
 /tests/pos/.liquid/
 /external/ocamlgraph/config.log
+/TAGS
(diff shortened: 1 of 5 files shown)


ranjitjhala:
Wow, thanks for figuring this out!

PR #224, opened 2016-05-03
Higher order quals (optionally) and currying on logical functions

nikivazou:


ranjitjhala:
Do LH tests pass? Are all changed "optional"?  any performance hits?

nikivazou:
LH tests pass

One change is optional: to use quals with function arguments (which leads to very bad performance)
The other edit (curry used defined logical functions) is not optional.

gridaphobe:
What about performance implications?

nikivazou:
Was just asking @christetreault how to easily compare performances

ranjitjhala:
Are the performance comparisons done here? (where are the CSV files?)

PR #225, opened 2016-05-05
Bound Eliminate

ranjitjhala:
Add a notion of "depth bound" to avoid eliminate blowing up. When run with `--elimBound=N` we DO NOT eliminate any KVar that is more than `N` (transitive) dependencies away another un-eliminated KVar. Yes its tricky. Paper forthcoming... :P


gridaphobe:
Is there a reasonable default for `elimBound`?

ranjitjhala:
`Nothing` (i.e. infinity). Except that it kills exactly one benchmark 
`Data/Text/Fusion.hs`. 

Specifically it kills a single function `mapAccumL` (where the bug 
was found, oddly...) which is not even a particularly "large" function...

This param is only used *if* we run `--eliminate`, so the default behavior is unchanged.

I expect a bit more head scratching is needed here before `--eliminate` 
can become the default, but getting very close...

gridaphobe:
I'm not sure what you mean by "killing" a benchmark, but I suspect you mean it times out, which is not a great default. That's why I was asking if there's a reasonable (non-infinite) default :)

ranjitjhala:
`12` works (i.e. doesn't "kill") for all benchmarks,
but makes some (many) slower than infinite...

​

PR #227, opened 2016-05-15
axiomatize function equality

nikivazou:


src/Language/Fixpoint/Types/Names.hs
```diff
@@ -44,6 +44,7 @@ module Language.Fixpoint.Types.Names (
   , dropSym
   , headSym
   , lengthSym
+  , mappendSym
```
gridaphobe: I believe we don't export this on purpose. Constructing a `Symbol` involves hashing the string and performing a table lookup, which is expensive, so we want to keep symbol construction to a minimum. Better to perform all formatting operations on the raw `Text` and then make a single `Symbol` at the end.

ranjitjhala: Yes please don't mess with that API ...

ranjitjhala: Why is mappend being exported? There are various other exported functions
that add suffixes etc. please use those instead!

nikivazou: Well, when ghc has two different type class with same names and "similar" arguments does a weird thing of just appending names to them. Following ghc given the symbol `X` I want to create the symbol `X0` and `X1` (without any separator).  I will define an intSymbolBare inside `Names` is this is any better.

nikivazou: Yes, this code https://github.com/ucsd-progsys/liquidhaskell/blob/f399e197dec7bc82575a2d2ce6be34bbc06283c4/benchmarks/haskell16/pos/Proves.hs

creates dictionaries 

`$fOptEqa(->)` and `$fOptEqa(->)0`

nikivazou: Because of line 129, 139

gridaphobe: I see, it has to do with overlapping instances. Why do you need to create the `Symbol` yourself then? Are you trying to lookup the dictionary `Id` by recreating the name? There must be a better way.

nikivazou: Yes, I am creating the symbol from the refined annotation `instance ToProve a (Proof -> b) where`. Let me know if you find a better way:)

ranjitjhala: Regardless, do not expose mappendSym, you can use symbolSuffix instead...

nikivazou: @ranjitjhala I cannot because symbolSyffix inserts a separator. I created another function in Names that does what I want, and do nor expose mappendSym, if this makes any difference

ranjitjhala: No Niki please don't do that, the separator is there for a reason. what is
the problem with inserting the separator?

nikivazou: @ranjitjhala from the refined definition `instance ToProve a (Proof -> b) where` I want to create the dictionary variables `$fOptEqa(->)` and `$fOptEqa(->)0` to get the variable names ghc creates.

PR #228, opened 2016-05-23
fix a bug in Minimize.hs

bmcfluff:
The added test shows a case where the "unsat core" returned by Minimize is actually sat. I think this PR fixes the deltaDebug algorithm.

### DIFF
--- a/src/Language/Fixpoint/Minimize.hs
+++ b/src/Language/Fixpoint/Minimize.hs
@@ -58,7 +58,7 @@ deltaDebug testSet cfg solve finfo set r = do
             then deltaDebug testSet cfg solve finfo s2 r
             else do
               d1 <- deltaDebug testSet cfg solve finfo s1 (s2 ++ r)
-              d2 <- deltaDebug testSet cfg solve finfo s2 (s1 ++ r)
+              d2 <- deltaDebug testSet cfg solve finfo s2 (d1 ++ r)
               return (d1 ++ d2)
 
 testConstraints :: (NFData a, Fixpoint a) => Config -> Solver a -> FInfo a -> ConsList a -> IO Bool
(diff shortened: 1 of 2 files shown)


ranjitjhala:
@rolph-recto can you take a quick look at this?

rolph-recto:
Sorry for the late response, I've been away for family business.

@bmcfluff The reference for the delta debug implementation is Algorithm 1 of this paper: http://web2.cs.columbia.edu/~junfeng/09fa-e6998/papers/delta-debug.pdf

The problem with using delta debugging with constraints is that constraint sets are "ambiguous" (see Section 3 of the paper I linked), which means roughly that there might be more than one cause of failure in the set. (An unambiguous set is one where given two failing subsets A and B, the intersection of A and B is also failing.) The DD algorithm assumes the domain it's working on is unambiguous, which I think is the cause of the bug you found. Partitioning the set into connected components is a heuristic to take care of this, but as you can see it isn't perfect.

For what it's worth, here are the runs of the DD algorithm for the original implementation and (2) your modified implementation. * means that the constraint is included, - means that it isn't:

```
(1) Original
1 2 3 4 | Safe?
* * - - | Yes
- - * * | Yes
* - * * | No
* * * - | No
Minimal failing set: 1,3 (SAFE)

(2) Modified
1 2 3 4 | Safe?
* * - - | Yes
- - * * | Yes
* - * * | No
* - * - | Yes
* - - * | No
Minimal failing set: 1, 4 (UNSAFE)
```

As you can see, your modification was able to determine that 3 is not relevant since {1,3} is not a failing set, and then finds {1,4} is unsafe and thus the minimal failing set. The original fails to determine this because {1,3,4} and {1,2,3} are both unsafe and it assumes that the minimal failing set must be in the intersection of these two sets -- thus {1,3}. But since the set is ambiguous, {1,3} is actually not failing.

As far as your modification, I have to think about its correctness. It's intuitive why it works but I'm afraid there might be a counterexample that allows it to pass over constraints that are part of the minimal failing set. I think the more robust solution would be to check if the minimal failing set is indeed failing, and continue the search if it isn't.

bmcfluff:
In our ambiguous domain, there can be more than one minimal failing set. I'm fairly sure the modified algorithm does always find _a_ minimal failing set, though there's no way to make it pick the smallest one.

PR #233, opened 2016-05-30
Fast elim

ranjitjhala:
Misc. refactoring and shuffling things around (probably breaks clients like LH), before starting the "new" eliminate.


ranjitjhala:
(even `stack haddock` worked fine...)

PR #242, opened 2016-06-12
take makeContext's Bool from caller instead of always using False

zjhmale:
i think take the Bool value to control the [different semantics](https://github.com/ucsd-progsys/liquid-fixpoint/blob/master/src/Language/Fixpoint/Smt/Theories.hs#L116-L117) from caller outside is quite necessary, so maybe we should elininate this TODO via this PR?

other places depend on `mkZ3Context` should be update such as [here](https://github.com/ucsd-progsys/prover/blob/master/src/Prover/SMTInterface.hs#L9).

### DIFF
--- a/src/Language/Fixpoint/Smt/Interface.hs
+++ b/src/Language/Fixpoint/Smt/Interface.hs
@@ -97,10 +97,9 @@ runCommands cmds
 -}
 
 
--- TODO take makeContext's Bool from caller instead of always using False?
-makeZ3Context :: FilePath -> [(Symbol, Sort)] -> IO Context
-makeZ3Context f xts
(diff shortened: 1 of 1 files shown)


ranjitjhala:
Hi @zjhmale, 

many thanks for the PR! As it happens this was already fixed in `develop` (where we pass a `Config` instead of a `Bool`). In general, `master` is deliberately a bit old and "stable" while the latest fixes and PRs go into `develop`.

Thus, am closing this PR, but will be happy to take any others you may have!

Best!,

-Ranjit.

zjhmale:
@ranjitjhala that's fine, i'll check out the `develop` branch :)

PR #246, opened 2016-06-18
generalize numeric tycons

nikivazou:


nikivazou:
This allows Fractional (Numeric) instances to be (extentionally) treated as reals (ints) in logic. See [tests/pos/FractionalInstance.hs](https://github.com/ucsd-progsys/liquidhaskell/blob/03491e257e39c127f29a0810ead95e052bd96b5b/tests/pos/FractionalInstance.hs) for an example. 

Goes with PL [liquidhaskell/pull/739](https://github.com/ucsd-progsys/liquidhaskell/pull/739). 
All tests pass. Ready to merge!

PR #247, opened 2016-06-23
Lambda rerpesentation

nikivazou:
Lambda terms are represented as `smt_lam x e` with `smt_lam :: a -> b -> Fun a b` when extensionality flag is off


ranjitjhala:
Please merge whenever you are ready

PR #249, opened 2016-06-28
liquid-fixpoint now builds with GHC 8

christetreault:


nikivazou:
Great @christetreault! So, it builds with both GHC7 and 8 right?

christetreault:
According to CircleCI it builds with GHC 7.10.2. I just added some `#if`'s in there, so I can't imagine it wouldn't

gridaphobe:
Why are the instances not needed with GHC 8?

ranjitjhala:
Is there any reason not to merge this? (i.e. will it break for us 7.10 weaklings?)

christetreault:
Let's hold off on this one for the time being; these instances are probably used by liquid-haskell or something...

PR #259, opened 2016-07-13
String interpretation by Z3

nikivazou:
`--stringtheory` flag enables interpretation of string constants by z3 (with interpreted functions `str.len` and `str.substr`). 

NOTE: this requires the `Z3` installed from their github repo: current released versions do not support string interpretations.


gridaphobe:
Please add a version check like

https://github.com/ucsd-progsys/liquid-fixpoint/blob/7382aa3e765eb87234a6d6527aa1e1cef5535d7f/src/Language/Fixpoint/Smt/Interface.hs#L297-L315

so users with older Z3s (or other solvers) get a nice error :)

nikivazou:
Good point @gridaphobe, how do we print errors nicely in fixpoint?

PR #265, opened 2016-10-05
Infix Operators in the logic

nikivazou:
This PR adds infix operators in the logic. 

Parser now has a state, and during parsing infix operators can be defined modifying `buildExpressionParser`.


ranjitjhala:
Omg!

PR #270, opened 2016-11-03
removed all trailing whitespace

atondwal:
So we don't have to mix whitespace fixes added by git with semantic fixes in our patches (polluting our commit history)


gridaphobe:
While you're at it, can you ensure there are no tabs in the source code (probably already the case)?

atondwal:
Already did that! :)

PR #278, opened 2017-01-07
[WIP]  Added support for Z3 pattern triggers

atondwal:
Currently absent from the .fq files, but present in the .bfq as an extra
parameter to the PAll constructor in Expr.

Visitors ignore that parameter.

This is still a work-in-progress; PR just to track progress.


ranjitjhala:
Any features for fixpoint need to have accompanying test cases in human
readable (and editable) .fq files...

atondwal:
Right, working on those!

PR #279, opened 2017-01-08
Elaborate bug

nikivazou:
The sorts in the expressions were not elaborated, i.e., defuntionalized.


ranjitjhala:
Thanks! Can you check that all LH tests pass before merging though?

PR #281, opened 2017-01-15
Limit supported compilers to GHC 7.10.*

alanz:
As it does not build with any other versions currently

### DIFF
--- a/liquid-fixpoint.cabal
+++ b/liquid-fixpoint.cabal
@@ -50,7 +50,7 @@ Executable fixpoint
   if flag(devel)
     ghc-options: -Werror
   hs-source-dirs: bin
-  Build-Depends: base >= 4.7 && < 5
+  Build-Depends: base == 4.8.*
                , liquid-fixpoint
 
 Library
@@ -105,7 +105,7 @@ Library
                    Language.Fixpoint.Utils.Statistics,
                    Language.Fixpoint.Solver.Solve
 
-  Build-Depends: base >= 4.7 && < 5
+  Build-Depends: base == 4.8.*
                , array
                , async
                , attoparsec


gridaphobe:
What compiler are you using, and what is the error? 

LiquidHaskell is definitely limited to GHC 7.10 as it makes heavy use of the GHC's API, but liquid-fixpoint **ought** to work across versions. So I'd rather fix the build than constrict the version range.

alanz:
I was not sure about that, given it works with liquidhaskell.  With GHC 8.0.1 I get

```

src/Language/Fixpoint/Types/Errors.hs:67:10: error:
    • No instance for (Serialize
                         (Text.PrettyPrint.Annotated.HughesPJ.Doc ()))
        arising from a use of ‘Data.Serialize.$dmput’
      There are instances for similar types:
        instance Serialize Doc
          -- Defined at src/Language/Fixpoint/Types/Errors.hs:67:10
    • In the expression: Data.Serialize.$dmput
      In an equation for ‘put’: put = Data.Serialize.$dmput
      In the instance declaration for ‘Serialize Doc’

src/Language/Fixpoint/Types/Errors.hs:67:10: error:
    • No instance for (Serialize
                         (Text.PrettyPrint.Annotated.HughesPJ.Doc ()))
        arising from a use of ‘Data.Serialize.$dmget’
      There are instances for similar types:
        instance Serialize Doc
          -- Defined at src/Language/Fixpoint/Types/Errors.hs:67:10
    • In the expression: Data.Serialize.$dmget
      In an equation for ‘get’: get = Data.Serialize.$dmget
      In the instance declaration for ‘Serialize Doc’
```

alanz:
And with GHC 7,8

```
Configuring component lib from liquid-fixpoint-0.5.0.1...
Preprocessing library liquid-fixpoint-0.5.0.1...

src/Language/Fixpoint/Solver/Graph.hs:1:14:
    Unsupported extension: PartialTypeSignatures
cabal: Failed to build liquid-fixpoint-0.5.0.1 (which is required by
exe:fixpoint from liquid-fixpoint-0.5.0.1).
```

gridaphobe:
Thanks! These both look easy to resolve :)

gridaphobe:
Thanks @alanz, I've fixed the GHC 8 incompatibility, and marked 7.8 as incompatible though (too much CPP to maintain it).

PR #282, opened 2017-01-16
additions for liquid instances

nikivazou:


ranjitjhala:
Thanks, doesn't look like this stuff is guarded by a command line / `Config` flag?

nikivazou:
This just exports some interface functions. 

Fixpoint does not know anything about Liquid Haskell instantiations.

Issue #283, opened 2017-01-19
Merge ghc8 branch


woehr:
Is there a reason the ghc8 branch hasn't been merged?

From what I can tell, it compiles fine on develop with ghc 7.10 and 8.0. I have noticed that the additional dependency on cereal-text wasn't added to the nix files and also that there is a warning that could be suppressed. I'd be happy to make a pull request for these minor issues.

ranjitjhala:
Hi @woehr --

I think the main reason is that:

1. An important "client" of `liquid-fixpoint` (`liquidhaskell`) is still a ways from being able 
    to work with GHC 8, and,

2. Looks like that branch has gotten a bit stale (i.e. is well behind `develop`)

It would be _awesome_ if you could update the `ghc8` branch to be upto-date 
with `develop` and make the above change so we can move `fixpoint` up to GHC 8.

Thanks!

woehr:
Unbeknownst to me, GHC8 support was added a few days ago (and in a way that doesn't add the cereal-text) dependency. However I can't build using the stack 7.16 resolver because of the located-base dependency being out of date. It seems though that located-base isn't used, and removing it fixes the ghc8 build and doesn't cause 7.10.3 to break. @gridaphobe, since you did the ghc8 update, and located-base is yours, could you comment?

gridaphobe:
What is the problem with located-base and GHC8? I thought I had pushed an update to hackage (0.1.1.0 IIRC).

woehr:
Liquid-fixpoint won't compile with stack+ghc8 because the located-base version hasn't been updated in stack.yaml.

But what I'm curious about is whether located-base is required since it doesn't seem to be used anywhere, and liquid-fixpoint will compile fine without it.

ranjitjhala:
If fixpoint compiles fine without it, lets just drop it!

(@gridaphobe, are we missing something here?)

gridaphobe:
At one point we were using it to add callstacks to the error/panic calls. There are a bunch of functions in L.F.Misc that optionally use it (eg `safeZip`).

So I would rather update it to 0.1.1.0 (which compiles against 7.10 and 8.0) and add it back to the .cabal file.

PR #284, opened 2017-01-20
Update located-base version in stack.yaml

woehr:
Addresses #283


gridaphobe:
Yep, thanks!

Btw, is there a convenient way to use multiple GHCs with stack?

PR #292, opened 2017-03-14
Gradual

nikivazou:
Adding support for Gradual Refinement Types.


src/Language/Fixpoint/Types/Solutions.hs
```diff
@@ -132,27 +162,37 @@ instance PPrint QBind where
 -- | A `Sol` contains the various indices needed to compute a solution,
 --   in particular, to compute `lhsPred` for any given constraint.
 --------------------------------------------------------------------------------
-data Sol a = Sol
+data Sol b a = Sol
```
ranjitjhala: Did you see my remark about this `b`? Do you _really_ need it?

nikivazou: Where is your remark? Yes, I do need it, because without Gradual Typing this is a `()` but with I use `b` to store the gradual solutions.

ranjitjhala:
@nikivazou -- odd I was about to rant that you hadn't made any of the changes from my message, but mysteriously I find that the message didn't even show up on github, even though it seems sent from gmail :-/ 

Anyways, here's the long email that I sent a few days ago but which seems to have gotten lost.

Can you please make these changes before merging? Thanks!

---

Here's a list of comments. (Please don't merge till addressed...)

0. The most important thing -- I don't see why you need to add a second parameter to Sol

https://github.com/ucsd-progsys/liquid-fixpoint/pull/292/files?diff=split#diff-1d67a72e608591bfed505af139e12244R165

  It seems like the `b` parameter is basically 
  just a list of the original `[a]`.

  So you should be able to do something like

```
    , gMap  :: !(M.HashMap KVar [a])         -- ^ Solution for gradual variables 
```

  or more precisely,

```
   , gMap  :: !(M.HashMap KVar (GBinder a))
```

  where 

```
    type GBinder a = (((Symbol, Sort), Expr), [a]) 
```

Then you don't have _two_ type parameters which are a real pain to work with. (You can just use an empty hash map for non-gradual mode? 

1. This whole thing should be cleaned up.

https://github.com/ucsd-progsys/liquid-fixpoint/pull/292/files?diff=split#diff-8ef29b0b018fccdc53d15fe23630d4b3R232

A. I do not like local functions used in this style
   `showGradual`, `showppfixed`, `ident` etc.; 
   pull them out as top level definitions.

B. "Guard" the output AFTER `showpp (resSolution res)`  with some sort of flag,
   otherwise it will print junk after the solution even in the "non-gradual"
   case.

2. This line is doing too many things, and hence, is not nice

https://github.com/ucsd-progsys/liquid-fixpoint/pull/292/files?diff=split#diff-bf01476b91500c411db7d0fadd6bc98fR76

split it into something like

```
  w'  = extraMagic fi  w''
  w'' = w { wenv = insertsIBindEnv newIds mempty, wrft = (v', t, k) }

-- pulled out into top-level function
extraMagic fi w = updateWfCExpr (subst su) w
  where 
    su          = mkSubst ((v, EVar v'):[(x, eVar $ kArgSymbol x (kv k)) | x <- kvarDomain fi k])
```
 

3. Please undo this (relations-on-bools) -- or give some reason why this extra 
   constraint (not-a-bool) is needed

https://github.com/ucsd-progsys/liquid-fixpoint/pull/292/files?diff=split#diff-ff210d38f74769efe444043cc0381991L786

   similarly, please undo this 

https://github.com/ucsd-progsys/liquid-fixpoint/pull/292/files?diff=split#diff-a0584b4b3cfb582e50e44f96471d365eL33

4. This formatting looks gross, 

https://github.com/ucsd-progsys/liquid-fixpoint/pull/292/files?diff=split#diff-1d67a72e608591bfed505af139e12244R26

make it something like

```
   -- * Solution elements		   
   , Hyp, Cube (..), QBind		
   , EQual (..)		
   , eQual, trueEqual

   -- * Gradual Solution elements		   
   , GBind, qbToGb, gbToQbs, gbEquals, equalsGb, emptyGMap
```

ranjitjhala:
Also no worries if you can't /don't do 0 -- the type parameter thing, your explanation makes sense and it will be easy to switch back to my approach later if needed.

nikivazou:
Thanks @ranjitjhala! I never saw this notes on github. Will address them tomorrow and let you know.

ranjitjhala:
Ok! Now I wonder what else hasn't been showing up on GH -- wonder if it was
because of the two-factor authentication thing ... oh well!

nikivazou:
@ranjitjhala I addressed all your things. 

Apart from 0 that you said it is ok to leave as is?

I did restore the bool comparison before these updates. But I still do not like it. I do not understand what is the meaning of `x < True` and why this is usefull. Note that when I added this check no Liquid Haskell benchmark failed...

ranjitjhala:
Thanks! Please go ahead and merge!

ranjitjhala:
Ok thanks! please merge (also can you ack if you get this message?)

nikivazou:
Got it! Thanks!

PR #297, opened 2017-03-20
Fix build for GHC 8.0.2

alanz:


ranjitjhala:
Many thanks @alanz !!!

PR #299, opened 2017-03-20
Clean up the parser (a bit)

alanz:
Adds a testsuite for the parser, and removes a number of `try` uses.

The main `liquidhaskell` tests also pass using this version, as seen [here](https://circleci.com/gh/alanz/liquidhaskell/26)


ranjitjhala:
OMG, this is amazing! THANKS @alanz !!!

PR #301, opened 2017-03-23
Matching changes for the parser-az branch in liquidhaskell

alanz:
The most important changes are

- make `reserved` and `reservedOp` work properly
- bring in the rest of the liquidhaskell reserved words


ranjitjhala:
🙇 🙇 thanks!

PR #303, opened 2017-04-09
Match changes for liquidhaskell lower case reserved words

alanz:
See https://github.com/ucsd-progsys/liquidhaskell/issues/971
and
https://github.com/ucsd-progsys/liquidhaskell/pull/985


ranjitjhala:
thanks!!!

PR #305, opened 2017-04-20
Symbolic Evauation

atondwal:
This patchset brings in the symbolic evaluation functionality from
LiquidHaskell into fixpoint. Still need to write plaintext fq parsing and
tests.

### DIFF
--- a/liquid-fixpoint.cabal
+++ b/liquid-fixpoint.cabal
@@ -103,6 +103,7 @@ Library
                    Language.Fixpoint.Solver.UniqifyKVars,
                    Language.Fixpoint.Solver.UniqifyBinds,
                    Language.Fixpoint.Solver.Eliminate,
+                   Language.Fixpoint.Solver.Instantiate,
                    Language.Fixpoint.Solver.Sanitize,
                    Language.Fixpoint.Utils.Statistics,
                    Language.Fixpoint.Solver.GradualSolve,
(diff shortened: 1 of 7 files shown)


ranjitjhala:
The `L.F.Types.Constraints` is getting a bit big; can you put `AxiomEnv` and related in a separate module/file? Thanks!

nikivazou:
Shouldn't this have a respective LH pull request that 
- reomoves all the evaluation code from Liquid Haskell 
- uses this fixpoint code instead
- while all the tests still pass?

ranjitjhala:
good point...

atondwal:
Here's that PR. N.B. Circle won't be able to build or test that PR until this PR is merged
https://github.com/ucsd-progsys/liquidhaskell/pull/993

atondwal:
> can you put AxiomEnv and related in a separate module/file?

Hmm, I'm not sure how I should do that `FInfo` contains an `AxiomEnv`, which depends on `SubcId` exported from `L.F.T.Constraints`, so if I made an `L.F.T.AxiomEnv` and put it there, we'd have a circular dependency. We could solve that by moving `FInfo` into its own file too, but that seems like a pretty big change; do we want to do that?

ranjitjhala:
ok, lets merge this for now and can fiddle with the AxiomEnv etc. later!

@atondwal I will let you / @nikivazou  merge this in case there are synch issues with LH.

ranjitjhala:
please don't wait for me to merge...

atondwal:
I don't have push access

ranjitjhala:
According to github @atondwal has push access...?

atondwal:
Oh, I'm so silly; I cloned over https instead of ssh, and was trying to do a `git push origin symeval:develop`, and obviously that didn't work :embarassed:

ranjitjhala:
Apparently in such situations one can just edit
the .git/config file and make the `url` field be `git@github.com...`

PR #311, opened 2017-05-03
Parsing and printing AxiomEnv

atondwal:
This adds support for instantiation from plaintext fq files.

### DIFF
--- a/src/Language/Fixpoint/Parse.hs
+++ b/src/Language/Fixpoint/Parse.hs
@@ -665,7 +665,16 @@ pairP xP sepP yP = (,) <$> xP <* sepP <*> yP
 --  where
 --    Just (c,cs) = T.uncons $ symbolSafeText s
 
+---------------------------------------------------------------------
+-- | Axioms for Symbolic Evaluation ---------------------------------
+---------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


ranjitjhala:
​Nice! Is there an example .fq file that shows what the format is?

or is this still WIP? ​

also, what is this FUEL?

atondwal:
Yep, it's

```
constant len: (func(1, [(Main.List  @(0)); int]))
constant Cons: (func(2, [@(0); (Main.List  @(0)); (Main.List @(0))]))
constant Nil: (Main.List @(0))

match len Nil = 0
match len Cons x xs = (1 + len xs)

constraint:
  env []
  lhs {v : int | true }
  rhs {v : int | len(Cons(1, Cons(2, Cons(3, Nil)))) = 3}
  id 1 tag []

expand [1 : True]
```

I'll push it into the tests dir after I add it to the test harness.

atondwal:
Here's the corresponding LH PR, since I made a minor change to the AEnv API: 
https://github.com/ucsd-progsys/liquidhaskell/pull/1017

nikivazou:
Why is this merged? It is not compatible with the LH develop

atondwal:
Here's the LH PR: https://github.com/ucsd-progsys/liquidhaskell/pull/1017. I want to merge it, but I'm worried about circle...

PR #312, opened 2017-05-03
Ghc 8

nikivazou:
This is what Liquid Haskell develop depends on, so better if this is the fixpoint develop.

### DIFF
--- a/liquid-fixpoint.cabal
+++ b/liquid-fixpoint.cabal
@@ -75,7 +75,6 @@ Library
                    Language.Fixpoint.Types.Triggers,
                    Language.Fixpoint.Types.Solutions,
                    Language.Fixpoint.Types.Utils,
-                   Language.Fixpoint.Types.Graduals,
                    Language.Fixpoint.Types,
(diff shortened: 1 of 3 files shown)


ranjitjhala:
@nikivazou -- I don't have a strong opinion either way, but in general we can keep the older dependency, and when ppl are building with LH, then cabal/stack should do the right thing,
right?

gridaphobe:
Yes, there's no reason to base fixpoint's requirements on LH's. Cabal/Stack will sort out the dependencies correctly.

nikivazou:
Ok! My main problem is that I am used to merge my local under-development branch to the current develop. 
But now that I did that, my gradual branch is not compatible with LH.

gridaphobe:
I don't follow, what precisely happens?

nikivazou:
Just that the current Fixpoint develop does not build with the current Liquid Haskell develop. 
Which I do not like, but not such a huge issue.

gridaphobe:
But why does it not build? I don't see how the change to the lower bound for `base` could affect the LF/LH compatibility.

PR #313, opened 2017-05-07
Gradual

nikivazou:


nikivazou:
All the edits are in the guarded gradual parts, so merging.

PR #317, opened 2017-08-02
Instantiate: instantiate measures before TopApps

atondwal:
This makes the following proof go through:

```haskell
{-@ LIQUID "--exact-data-con"                      @-}
{-@ LIQUID "--higherorder"                         @-}
{-@ LIQUID "--automatic-instances=liquidinstances" @-}

module Peano (minus) where

data Peano = Z | S {prev :: Peano} deriving (Eq)
{-@ data Peano = Z | S {prev :: Peano} @-}

{-@ measure minus @-}
minus :: Peano -> Peano 
minus (S n) = n 
minus Z     = Z 

{-@ test :: n:Peano -> m:Peano -> { v:() | S n == S m } -> { n == m } @-}
test :: Peano -> Peano -> () -> ()
test _ _ _ = ()
```

### DIFF
--- a/src/Language/Fixpoint/Solver/Instantiate.hs
+++ b/src/Language/Fixpoint/Solver/Instantiate.hs
@@ -289,7 +289,10 @@ assertSelectors γ e = do
     go (SMeasure f dc xs bd) e@(EApp _ _)
       | (EVar dc', es) <- splitEApp e
       , dc == dc', length xs == length es
-      = addSMTEquality γ (EApp (EVar f) e) (subst (mkSubst $ zip xs es) bd)
+      = let e1 = EApp (EVar f) e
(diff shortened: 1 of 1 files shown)


ranjitjhala:
We can (and should merge) this, but it will be made redundant by the `adt` stuff, see:

  https://github.com/ucsd-progsys/liquid-fixpoint/blob/adt/tests/pos/adt_list_1.fq

atondwal:
In this case, yes, but in general we should make sure we add all the measure information for all of the constructors the users actually use, eg if we want to use the fact that `size [] = 0`.

(Hmm, we should probably evaluate inside measures too, so that we can use the fact that `size [x] = 1`.)

ranjitjhala:
👍

PR #340, opened 2017-10-11
Add support for "equality constraints" in Sorts

ranjitjhala:
To support `(a ~ Int)` style equality constraints, 
as needed e.g. to support GADTs. 

https://github.com/ucsd-progsys/liquidhaskell/issues/1089

This works by generalizing `SortCheck.checkNumeric` to `checkEqConstr`

See: 

* tests/pos/EqConstr0.fq 
* tests/pos/EqConstr1.fq 
* tests/crash/EqConstr0.fq 
* tests/crash/EqConstr1.fq


ranjitjhala:
Some notes.

```haskell
λ> :i EntityField
class PersistEntity record where
  type role EntityField nominal nominal
  data family EntityField record a
  	-- Defined at ExactGADT3.hs:15:3
data instance EntityField Blob dog where
  BlobXVal :: EntityField Blob Int
  BlobYVal :: EntityField Blob Int
  	-- Defined at ExactGADT3.hs:35:8
```

So `EntityField Blob` is not a plain `TyCon` but a `data instance`  (type-family instance?)
of type `FamInst` (see `mgi_fam_inst`).

1. Nevertheless, `BlobXVal` and `BlobYVal` are `DataCon`s,
2. The family instance _does_ have a `TyCon` called `Query.R:EntityFieldBlobdog` 
3. But, but we cannot use the `TyCon` name directly as it kills the `lookupGhcTyCon` 
    (with a parse error) I guess its not supposed to be visible to the user.

**Solution (?)** extend `makeHaskellDataDecls` to also handle `FamInst` ?

PR #341, opened 2017-10-18
Export few fixpoint functions

nikivazou:


ranjitjhala:
Why do you need to export simplifyFInfo?

Why is develop broken?

PR #342, opened 2017-10-18
Minor edits: exporting functions & deleting old gradual solver

nikivazou:


ranjitjhala:
Are you using simplifyFInfo inside fixpoint (ok) or in LH (not ok)

nikivazou:
I am using it inside the `Gradual`, a Target-like executable that is a client of both LH (for constraint generation) and fixpoint (for constraint solving). 

The idea is that I am solving the LH generated constraints many times, for various "gradual solutions" so I want to do the simplification phase only once.

ranjitjhala:
Sounds good!

PR #343, opened 2017-10-21
Minor: add exprSort_maybe so that liquid does not die in case of failure

nikivazou:


ranjitjhala:
Good idea! Please merge (if tests pass etc)

PR #352, opened 2017-12-20
change coercions to coercions and casts

nikivazou:


ranjitjhala:
Nice! These changes are strictly an improvement [they seem to make everything simpler!]

If it works for  LH then please merge, thanks!!

PR #354, opened 2018-01-31
Make CoSub visible externally

ranjitjhala:
We need it if we have ty-vars floating around inside refinements, e.g. in `ELam` and `ECoerc` because as TyVars are substituted, we need to also update the refinements. Matching PR for:

https://github.com/ucsd-progsys/liquidhaskell/pull/1226


ranjitjhala:
Also see:

https://github.com/ucsd-progsys/liquidhaskell/pull/1232

PR #355, opened 2018-02-18
Parsing support for singleton and empty lists in the refinements

nikivazou:
See https://github.com/ucsd-progsys/liquidhaskell/blob/d1b94e27043d8cc049fdfb2958a741465e75f252/tests/pos/SingletonLists.hs

matches 
https://github.com/ucsd-progsys/liquidhaskell/pull/1240


ranjitjhala:
thanks!!

PR #367, opened 2018-05-24
Elim ebinds

atondwal:
Precomputing the `EbSol`s before we refineKs is unsound, so this fixes that by recomputing them.


ranjitjhala:
Awesome so all the tests work now?

atondwal:
No, this is just for soundness --- in fact we compute strictly weaker solutions than before. I shou I should add negative test cases.

atondwal:
Okay, now all the LH tests do actually pass. (see matching LH PR https://github.com/ucsd-progsys/liquidhaskell/pull/1320)

Still not really sure why we're failing the tests in todo/ebind-05.fq --- we should be skipping the constraint that defines `v=5`, but in fact that's the only constraint that hits z3.

I also tried to port over some of tachio's examples, but I realized that a lot of them are dependent on recursion, so that's sort of out.

ranjitjhala:
As a general rule, I don't like to merge things till
we understand why things work/fail, so can you see if
you can decipher the `ebind-05.fq` else can discuss next
week.

Good idea re: Tachio examples; but no need to give up
so soon. What do, say, 3 of the _simplest_ tests look
like (directly ported to Haskell)? We can figure out
if the recursion can be "factored out".

atondwal:
Ah, figured it out ---  it's that for some ebind `n`, `isConc [| v= n |]` is `True`, so even if we eliminate it in the graph, it can still get added to `wConc`s at the "end" of the worklist, and thrown at z3.

> Good idea re: Tachio examples; but no need to give up
so soon. What do, say, 3 of the _simplest_ tests look
like (directly ported to Haskell)? We can figure out
if the recursion can be "factored out".

I've already ported about half of Tachio's examples to Haskell, and the 3-4 simplest ones do go through (once I added the appropriate type annotations). I'll make a `benchmarks` dir in the matching LH PR and add both the working and nonworking ones after I port the rest of them. We can surely get a few more to go through, but it would've been really nice to get the two big running examples from his paper working, though they're running examples precisely because they show how/that he deals with the added complexity due to recursion.

PR #378, opened 2018-11-03
Eb inhabation

atondwal:
WIPish. Just need to make it emit an appropriate error message and then it'll be ready for merge.


ranjitjhala:
Excellent! May be good to add a neg test to check that we can fail?

atondwal:
For the lack of a better constraint to blame, I'm just going to blame all of the defining constraints

PR #384, opened 2019-01-09
Upgrade to GHC 8.4.4.

mhwombat:
These changes will allow liquid-fixpoint to build with GHC 8.4.4. I have tested building it with both Stack and Nix/NixOS. You may not be ready for that yet, so feel free to disregard this pull request.

There are a lot of changes to default.nix and shell.nix. That's because I'm using the versions that are automatically generated by `cabal2nix`. I think this will be simpler for others who might work on the package.

For future reference, here's how I generated those files.

```
cabal2nix . > default.nix
cabal2nix . --shell > shell.nix
```

I am creating a similar pull request for liquidhaskell.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,25 +1,25 @@
 { fetchgitLocal }:
 { mkDerivation, ansi-terminal, array, ascii-progress, async
 , attoparsec, base, bifunctors, binary, boxes, bytestring, cereal
-, cmdargs, containers, deepseq, directory, filemanip, filepath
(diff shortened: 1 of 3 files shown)


peti:
I think it's safe to say that this PR can be closed. Nix has moved on to GHC 8.6.x a while ago. Generally speaking, I believe that `default.nix` and `shell.nix` should be removed from this repository entirely because these are generated files that carry no actual information. Anyone who wants them can easily generate them with `cabal2nix`.

PR #393, opened 2019-03-28
drop redundant dependencies from Cabal file

peti:
17e752c is where the substance is. The second commit just makes the file prettier without actually changing any of the meaning.


nikivazou:
Thanks! 
It would be great if you can link this to LH with a resp. PR. But maybe you could wait to merge my current PR first to avoid conflicts (since submodule conflicts are weird...)

PR #399, opened 2019-05-09
wait for SMT to write before you read

nikivazou:
addresses oef SMT exceptions


ranjitjhala:
Hmm why do we need this extra check?
Shouldn’t getline call block until the data is available?

nikivazou:
It should! 

But, I have a file with a very slow SMT query and locally it crashes with 
```
liquid: fd:15: hGetLine: end of file
```
without this edit. 
There is a change in the implementation of `getline`. In the changeLog of `text` there is the following:

```
### 1.2.3.1

* Make `decodeUtf8With` fail explicitly for unsupported non-BMP
  replacement characters instead silent undefined behaviour (gh-213)

* Fix termination condition for file reads via `Data.Text.IO`
  operations (gh-223)
```

So, maybe they messed it up and I should file an issue there...

nikivazou:
Done! So, either we merge or we wait for this https://github.com/haskell/text/issues/258

ranjitjhala:
I assume the LH tests pass? Can you check it doesn’t slow things down (as this is in the critical path in the sky interface ... shouldn’t matter but may be good to check)

nikivazou:
According to the text people `hGetLine` does not block (and it is not supposed to). 
I am going to link this to LH to see if tests pass. 

I agree that there might be a slowdown, but it is better than SMT crashing.

ranjitjhala:
@nikivazou did you get a chance to re-run the LH benchmarks with this, btw?

nikivazou:
No, I just created the relevant LH PR, but did not time the benchmarks

ranjitjhala:
Ok, please run the benchmarks -- I don't want to merge this without doing so
as it could make _all_ programs much slower to put in a this check at that
point (or it most likely won't but its good practice to confirm
this point before merging.)

PR #405, opened 2019-06-13
Allow Tuple Syntax in Refinements

zgrannan:
This pull request allows for tuple syntax to be used in refinements. It can handle the example in https://github.com/ucsd-progsys/liquidhaskell/issues/1162.

However, this will change is not entirely backwards compatible. Currently, `f (x, y)` in a refinement is parsed as `f x y`. This syntax will now mean the same as it does in Haskell (i.e applying the single argument `(x, y)` to `f`). It looks like there are some tests in the `liquidhaskell` repository that indeed do break with this change.

### DIFF
--- a/src/Language/Fixpoint/Parse.hs
+++ b/src/Language/Fixpoint/Parse.hs
@@ -99,7 +99,6 @@ import           Language.Fixpoint.Smt.Bitvector
 import           Language.Fixpoint.Types.Errors
 import qualified Language.Fixpoint.Misc      as Misc      
 import           Language.Fixpoint.Smt.Types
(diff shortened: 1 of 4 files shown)


ranjitjhala:
Wow, this is awesome @zgrannan! 

Yes, some tests will break but I think we can/should update 
those because we really ought to parse tuples as tuples. 

Can you submit a matching PR in the LH repo that just links 
the submodule to this one, and then we can see which tests 
break and fix those and merge in?

Many thanks!!!

nikivazou:
Thanks a lot!!!! 

Apparently circle does not get activated for forks, but I run the tests locally and they pass.
So merging (but would be great to activate circle for PRs from forks too)

PR #417, opened 2019-12-14
Expand Function Extensionality

nikivazou:
When two function expressions, say `e1` and `e2` are compared on the RHS of a constraint they are saturated with a fresh argument `x` (e1 == e2 turns into e1 x == e2 x). 

See https://github.com/ucsd-progsys/liquid-fixpoint/blob/698f5f7a00bd3e70080c76538bf8eca7af6cd20f/tests/pos/T416.fq that is now OK.

### DIFF
--- a/liquid-fixpoint.cabal
+++ b/liquid-fixpoint.cabal
@@ -74,6 +74,7 @@ library
                     Language.Fixpoint.Solver.UniqifyBinds
                     Language.Fixpoint.Solver.UniqifyKVars
                     Language.Fixpoint.Solver.Worklist
+                    Language.Fixpoint.Solver.Extensionality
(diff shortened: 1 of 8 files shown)


ranjitjhala:
Is the “x” added as a fresh binder to the environment? (If not where does it come from?)

nikivazou:
Yes, the `x`s are accumulated and then added in the `bs` (https://github.com/ucsd-progsys/liquid-fixpoint/blob/698f5f7a00bd3e70080c76538bf8eca7af6cd20f/src/Language/Fixpoint/Solver/Extensionality.hs#L27)

ranjitjhala:
👍

PR #420, opened 2020-02-09
check whether binder is free in eliminateEta

yiyunliu:


nikivazou:
Thanks! Can you create a LH pull request that links this?

PR #421, opened 2020-02-14
fix for LH #1613

nikivazou:
Serialization of data declarations was ignoring the type parameters of the type constructions when serializing Type constructors.


ranjitjhala:
Thanks Niki!!! Pretty surprising this was working at all ... wonder if it’s
some change in smtlib?

nikivazou:
I don't think there is a change.
I assume it just happened and all the arguments were encoded as SMT ints.

PR #422, opened 2020-02-26
Get ple candidates from argument of applications

nikivazou:
Fix for https://github.com/ucsd-progsys/liquidhaskell/issues/1620

### DIFF
--- a/src/Language/Fixpoint/Solver/Instantiate.hs
+++ b/src/Language/Fixpoint/Solver/Instantiate.hs
@@ -412,7 +412,7 @@ topApps = go
     go (EBin  _ e1 e2) = go e1  ++ go e2
     go (PNot e)        = go e
     go (ENeg e)        = go e
-    go e@(EApp e1 e2)  = e:(go e1 ++ go e2)
+    go e@(EApp e1 e2)  = e:(concatMap go es) where (_,es) = splitEApp e
     go _               = []
 
 -- makeLam is the adjoint of splitEApp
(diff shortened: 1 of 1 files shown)


ranjitjhala:
@nikivazou   What is the motivation for this? The LH issue is too long :-) 

Please *do not* merge this without a simple .fq example that shows why the feature is needed.

(I really don't want to add things to PLE without very good justification.)

Is it that we were not, for example, doing PLE for function-arguments like, so e.g. given defs:

```
define bar(x) = 10
define foo(z) = z + 1
```

we could not prove that:

```
foo(bar 10) == 11
```

Or something like that? Either way, please add a test to the .fq

ranjitjhala:
For example, see

https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/tests/proof/ple4.fq

there needs to be a similar test for whatever this edit is doing.

nikivazou:
It is exactly what you described. 
What can we do, they are abusing ple with higher order functions... 
ok, let me first check if it breaks anything and then I will add the example

nikivazou:
BTW, do you have any recollection why we do not expand the arguments?

ranjitjhala:
Not really, am a bit surprised the various examples worked _without_
expanding the arguments TBH!

nikivazou:
> Not really, am a bit surprised the various examples worked _without_ expanding the arguments TBH!

The magic of congruence :) 

So, when this expansion is on, we end up triggering many partial applied functions, which is dangerous. E.g., we might expand partially applied data constructors, which are interpreted by the SMT, and then SMT complains.

PR #423, opened 2020-03-12
fix substitution bug in PLE

nikivazou:
For an equation `geq x y = if x >= y then true else false` the call `geq y x` would return 
```
(if x >= y then true else false) [x |-> y, y |-> x]
(if y >= y then true else false) [y |-> x]
(if x >= x then true else false) 
true
```

Solution: uniquefy the arguments in Equations. 

PS: The fact that we have to write `geq x y = if x >= y then true else false` instead of `geq x y = x >= y`  was what I was really trying to address and it is coming next :)


ranjitjhala:
Nice! Minor issue -- when you write

  the call `g y x` would return...

you mean

  the call `geq y x` would return ... right?

ranjitjhala:
Let me take a look a bit later today?

nikivazou:
Sure! 
I am digging into a deeper issue, so will make another PR, hopefully soon.

PR #426, opened 2020-03-18
New PLE Implementation

nikivazou:
This is a complete rewrite of PLE the follows more closely the theory presented in the POPL paper. 

COMPATIBILITY
----------------
Currently the new implementation is the default. You can restore the previous PLE with the `oldple` flag. The only incompatible different is that the `definition` parser has changed from 

```
define f (x:sx,y:sy):s = (f x y = e)
```

to 

```
define f (x:sx,y:sy):s = { e }
```

PROS
---------
0. It handles examples not supported before, e.g., [this](https://github.com/ucsd-progsys/liquid-fixpoint/blob/fb4e4b24c41932af9b725f1f7f14ce2052743848/tests/proof/pleBoolC.fq).
1.  All the test (including LH) apart from one (that is using lambdas) pass.
2.  PLE runs in all subexpressions (including arguments of functions which we need for some benchamarks) apart from subexpressions guarded with `EIte`. 
4. There is a small speedup in some examples (e.g., `Tests/Prover/foundations/Basics.hs`and critically the NN example under development).  
5.  It removes most hacks that existed before (there are two hacks discussed bellow*). Importantly: 
   - It removes the STACK FUN hack and is actually using SMT to check branches before entering EIte (as discussed in the paper). 
   - Function definitions are now actual definitions, e.g., `define f (x, y) = { e }` is now not treated as `f (x,y) = e`. Not sure how things worked with this previous encoding :) 
6. I now mostly understand what is happening and can edit the code when needed :) 

CONS
------
- There is a slow down is other examples (e.g., computation of `fib 16` is now terrible; but I will investigate more).
- I do not think I am taking advantage of the "incremental" implementation, since I do not understand how it worked. I left the `Trie` parts of the code, but do not seem to help much.  

Hacks/Departures from POPL paper*
-------
Evaluation proceeds as in the POPL paper, i.e., and SMT when there is an `EIte`, unfold defined functions, and recurse. Apart from the four speedup "hacks" below
- [cashed expressions](https://github.com/ucsd-progsys/liquid-fixpoint/blob/fb4e4b24c41932af9b725f1f7f14ce2052743848/src/Language/Fixpoint/Solver/PLE.hs#L363): subexpressions that are evaluated before are not evaluated again.   
- [constants](https://github.com/ucsd-progsys/liquid-fixpoint/blob/fb4e4b24c41932af9b725f1f7f14ce2052743848/src/Language/Fixpoint/Solver/PLE.hs#L601-L605): subexpressions known to evaluate to constants are directly simplified to such constants.
- Datatypes: There is special simplification treatment for [record selectors](https://github.com/ucsd-progsys/liquid-fixpoint/blob/fb4e4b24c41932af9b725f1f7f14ce2052743848/src/Language/Fixpoint/Solver/PLE.hs#L611-L615) and [measure](https://github.com/ucsd-progsys/liquid-fixpoint/blob/fb4e4b24c41932af9b725f1f7f14ce2052743848/src/Language/Fixpoint/Solver/PLE.hs#L599-L600), i.e., rewrite rules that reduce to constants. I bet both these can go away if we actually used SMT's data types instead of rewrite rules. 
- GADTs: To make examples with GADTs work we need to more hacks 
  1. [Remove trivial casts](https://github.com/ucsd-progsys/liquid-fixpoint/blob/fb4e4b24c41932af9b725f1f7f14ce2052743848/src/Language/Fixpoint/Solver/PLE.hs#L609-L610) to allow evaluation proceed. 
  2. [Simplify EIte when possible](https://github.com/ucsd-progsys/liquid-fixpoint/blob/fb4e4b24c41932af9b725f1f7f14ce2052743848/src/Language/Fixpoint/Solver/PLE.hs#L606-L608) so that malformed definitions (e.g, bool in one branch and int in the other) do not end up in SMT.


ranjitjhala:
Super!!!! will take a look, but please keep old code around just in case..

nikivazou:
@ranjitjhala all the old code is kept.
It would be good to have a PLE code review session, but in your own time.  
I will merge soon (about 2 hours) esp since I am seeing develop getting edited (😠  @atondwal )

ranjitjhala:
Go ahead Niki!

Was there a commit to develop without a PR?

PR #429, opened 2020-04-12
SMT-LIB datatypes

nikivazou:
Update to SMT-LIB datatype declaration syntax (which requires type constructors to be applied to type arguments in definitions). 

Also update circle z3 to z3-4.8 to support this syntax.


ranjitjhala:
Awesome!!! please

1 merge of tests pass
2 bump the fix point version on cabal to eflect change in z3 dependency

PR #435, opened 2020-06-03
Required extensions to support source plugins

adinapoli:
This commit extends `liquid-fixpoint` to fully support the new LiquidHaskell source plugin.

Most of the changes are innocuous, as we mostly add new instances (like `Hashable`) for some types. The only potential "breaking" change is the use of `ExitFailure 1` for `FixResult` even in case of validation errors, but as long as users didn't rely on a particular exit code in their programs, this should be fine.

I will annotate the PR with in-line comments for ease of review.

### DIFF
--- a/src/Language/Fixpoint/Solver/Monad.hs
+++ b/src/Language/Fixpoint/Solver/Monad.hs
@@ -28,8 +28,6 @@ module Language.Fixpoint.Solver.Monad
        )
        where
 
-import           Control.DeepSeq
-import           GHC.Generics
 import           Language.Fixpoint.Utils.Progress
 import qualified Language.Fixpoint.Types.Config  as C
 import           Language.Fixpoint.Types.Config  (Config)
@@ -37,19 +35,18 @@ import qualified Language.Fixpoint.Types   as F
 -- import qualified Language.Fixpoint.Misc    as Misc
 -- import           Language.Fixpoint.SortCheck
 import qualified Language.Fixpoint.Types.Solutions as F
-import           Language.Fixpoint.Types   (pprint)
 -- import qualified Language.Fixpoint.Types.Errors  as E
 import           Language.Fixpoint.Smt.Serialize ()
 import           Language.Fixpoint.Types.PrettyPrint ()
 import           Language.Fixpoint.Smt.Interface
 -- import qualified Language.Fixpoint.Smt.Theories as Thy
 import           Language.Fixpoint.Solver.Sanitize
+import           Language.Fixpoint.Solver.Stats
 import           Language.Fixpoint.Graph.Types (SolverInfo (..))
 -- import           Language.Fixpoint.Solver.Solution
 -- import           Data.Maybe           (catMaybes)
 import           Data.List            (partition)
 -- import           Data.Char            (isUpper)
-import           Text.PrettyPrint.HughesPJ (text)
 import           Control.Monad.State.Strict
 import qualified Data.HashMap.Strict as M
 import           Data.Maybe (catMaybes)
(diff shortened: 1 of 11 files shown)


src/Language/Fixpoint/Solver/Monad.hs
```diff
@@ -67,30 +64,11 @@ data SolverState = SS
   , ssStats   :: !Stats            -- ^ Solver Statistics
   }
 
-data Stats = Stats
```
adinapoli, line 70: We moved this `Stats` data structure into a standalone module to avoid cyclic imports, as this is now used inside `FixResult`.

unix/Language/Fixpoint/Utils/Progress.hs
```diff
@@ -18,12 +18,16 @@ pbRef :: IORef (Maybe ProgressBar)
 pbRef = unsafePerformIO (newIORef Nothing)
 
 withProgress :: Int -> IO a -> IO a
-withProgress n act = displayConsoleRegions $ do
-  -- putStrLn $ "withProgress: " ++ show n
-  progressInit n
-  r <- act
-  progressClose
-  return r
+withProgress n act = do
```
adinapoli, line 21: This change will effectively stop printing the progress bar for a verbosity of `Normal`. We did this to make sure that when Liquid is running via the plugin, we would get only the required amount of information, without too much noise, as this will now happen in lockstep with normal GHC compilation.

ranjitjhala: Is there some simple way to get a `--verbosity=progress` or similar? I think the two choices were `Normal` and `Loud` -- I'd like there to be some "middle" ground where we only see the progress bar? Alternatively, can we have a `--silent` which has no output at all, and run the PLUGIN under `--silent`?

ranjitjhala: Can we change the below to 

```
loud <- (Quiet /=) <$> getVerbosity
```

and run the plugin with `Quiet` verbosity? 
http://hackage.haskell.org/package/cmdargs-0.10.20/docs/System-Console-CmdArgs-Verbosity.html

adinapoli: Done as part of https://github.com/ucsd-progsys/liquid-fixpoint/pull/435/commits/a4e303bc55e8f596893e661954e20c983a61215f (see also https://github.com/ucsd-progsys/liquidhaskell/pull/1683/commits/9d6e7e879a4e2720cbfa2a1f52389c1978303d5b) . We now have a new `loggingVerbosity` field in the `Config` as opposed to the "opinionated" built-in `verbosity` (which doesn't account for a `--normal` flag).

Now `Quiet` is the new default (which the source plugin will use) but I have customised the executable to be passed the `--normal` flag "under the hood", so that we would log with the same level of verbosity (which will include also things like the progress bar).

As I was there, I have also removed the now-redundant `Serialize` instances for `Config` and `STMSolver`. Last but not least: while this commit is good enough for the time being, at some point we might want to revisit how to do logging, especially because now we have a situation where a LH's contributor could _still_ use the `setVerbosity` and `getVerbosity` functions from `cmdargs` which would essentially completely eschew the `loggingVerbosity` stored in the `Config` (i.e. the `Config` is not the single source of truth when it comes to logging). 

I think that in practice this is fine as the `Config` here only gives an initial "recipe" about what things should be. If we really wanted to do this properly we would have to somehow thread the `Config` all around and stop using the global functions from `cmdargs`, but that would require a significant sweeping change in the codebase, so it's probably not worth it for now (or possibly ever, 😸 )

PR #437, opened 2020-07-23
Update rewriting logic

zgrannan:
Improvements to rewriting:

- Rewriting is now more reliable (it now will apply rewrites to subexpressions)
- A termination checker for rewriting has been added (for details see: https://github.com/zgrannan/rest/blob/master/REST-POPL21.pdf)
- Added the option `--max-rw-ordering-constraints`. Lower values can speed up verification but may prevent application of rewrite rules.
- Added the option `--rw-termination-check`. This enables the rewrite termination checker (may slow down verification)
Also see LiquidHaskell main repo: https://github.com/ucsd-progsys/liquidhaskell/pull/1718

### DIFF
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -637,10 +637,9 @@ knowledge cfg ctx si = KN
   , knConsts                   = Mb.catMaybes $ map makeCons sims 
   , knAutoRWs                  = aenvAutoRW aenv
   , knRWTerminationOpts        =
-      if (noRWTerminationCheck cfg)
-      then RWTerminationCheckDisabled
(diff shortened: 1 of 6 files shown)


src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -198,7 +195,7 @@ data InstEnv a = InstEnv
   , ieSMT   :: !SMT.Context
   , ieBEnv  :: !BindEnv
   , ieAenv  :: !AxiomEnv 
-  , ieCstrs :: !(M.HashMap SubcId (SimpC a))
+  , ieCstrs :: ![(SubcId, SimpC a)]
```
zgrannan, line 198: Switch back to map

zgrannan:
TODO: Disable termination check by default

PR #440, opened 2020-08-19
fix test.hs to use fixpoint executable built from current source tree

plredmond:
Before

* test issues
    * `test.hs` uses `getBinDir` to locate `fixpoint` which is something like `/home/username/.cabal/bin` on my system. You need to do  `cabal v2-install` to put `fixpoint` in the directory before the tests will pass.
* nix issues
    * `default.nix` is statically generated, 5 years outdated, and doesn't pin the versions of the compiler or dependencies.

After

* test fixes
    * `test.hs` assumes that `fixpoint` can be found in `$PATH`
    * `liquid-fixpoint.cabal` uses `build-tool-depends:` to specify that `test.hs` needs `fixpoint` when `cabal v2-test` is run.
    * `default.nix` uses `$PATH` to ensure `fixpoint` is available when `nix-build` runs tests.
* nix fixes
    * `default.nix` dynamically generates its configuration from the cabalfile
    * `default.nix` pins versions of nixpkgs, all-cabal-hashes, ghc, and all dependencies for reproducibility.
    * `default.nix` patches an outdated package, `git-0.3.0` which isn't compatible with the removal of `fail` from `Monad` in ghc 8.10.1

Unknown

* I don't know how this will affect users of `stack`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ Session.vim
 /external/ocamlgraph/config.log
 /TAGS
 /tags
+*.swp
--- a/default.nix
+++ b/default.nix
@@ -1,32 +1,52 @@
-{ fetchgitLocal }:
-{ mkDerivation, ansi-terminal, array, ascii-progress, async
-, attoparsec, base, bifunctors, binary, boxes, bytestring, cereal
-, cmdargs, containers, deepseq, directory, filemanip, filepath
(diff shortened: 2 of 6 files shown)


plredmond:
Note, I'm completely replacing `default.nix` in this change because the existing one was generated by `cabal2nix` 5 years ago and most of the time these days you dynamically call `cabal2nix` as part of the build process (as the new `default.nix` does).

PR #442, opened 2020-08-24
Implement constant folding in PLE

zgrannan:
This results in a significant improvement in the speed of this test (~60s to ~4s):

https://github.com/ucsd-progsys/liquidhaskell/blob/develop/benchmarks/popl18/ple/pos/Fibonacci.hs

Corresponding PR in main repo: https://github.com/ucsd-progsys/liquidhaskell/pull/1734


ranjitjhala:
Thanks @zgrannan ! I'm presuming this didn't add any performance *regressions* ?

zgrannan:
> Thanks @zgrannan ! I'm presuming this didn't add any performance _regressions_ ?

I wasn't able to find any case where there was a performance regression.

PR #444, opened 2020-09-16
Rewrite and simplify the paser; add proper layout.

kosmikus:
[matches PR https://github.com/ucsd-progsys/liquidhaskell/pull/1704]

This is a major restructuring, even though more work remains to do.

The most important changes are:

- Use megaparsec instead of parsec.
- Change the handling of internal parser state.
- Improve error reporting somewhat.
- Improve location tracking somewhat.
- Add layout information to the parser state.
- Export layout block parser combinators.
- Add more coments and TODOs.

### DIFF
--- a/src/Language/Fixpoint/Parse.hs
+++ b/src/Language/Fixpoint/Parse.hs
@@ -15,32 +15,33 @@ module Language.Fixpoint.Parse (
   -- * Top Level Class for Parseable Values
   , Parser
 
-  -- * Lexer to add new tokens
-  , lexer
-
   -- * Some Important keyword and parsers
   , reserved, reservedOp
+  , locReserved
   , parens  , brackets, angles, braces
   , semi    , comma
   , colon   , dcolon
-  , whiteSpace
-  , blanks
+  , dot
   , pairP
   , stringLiteral
+  , locStringLiteral
 
   -- * Parsing basic entities
 
   --   fTyConP  -- Type constructors
   , lowerIdP    -- Lower-case identifiers
   , upperIdP    -- Upper-case identifiers
-  , infixIdP    -- String Haskell infix Id
+  -- , infixIdP    -- String Haskell infix Id
   , symbolP     -- Arbitrary Symbols
+  , locSymbolP
   , constantP   -- (Integer) Constants
-  , integer     -- Integer
+  , natural     -- Non-negative integer
+  , locNatural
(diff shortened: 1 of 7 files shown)


src/Language/Fixpoint/Parse.hs
```diff
@@ -579,18 +960,19 @@ sortFunP
...
 sortP' :: Parser [Sort] -> Parser Sort
 sortP' appArgsP
-   =  parens sortP
-  <|> (reserved "func" >> funcSortP)
-  <|> (fAppTC listFTyCon . single <$> brackets sortP)
-  <|> bvSortP
+   =  parens sortP -- parenthesised sort, starts with "("
+  <|> (reserved "func" >> funcSortP) -- function sort, starts with "func"
+  <|> (fAppTC listFTyCon . pure <$> brackets sortP)
+  <|> bvSortP -- Andres: this looks unreachable, as it starts with "("
```
ranjitjhala: hmm that's strange, but we do have some tests for it, search for `(BitVec Size32)` in tests/ so if they pass it should be ok?

ranjitjhala (approved):

ranjitjhala:
Hi @kosmikus -- this is great, shall we merge? Was there a matching PR for `liquidhaskell` too?

src/Language/Fixpoint/Parse.hs
kosmikus: I think `bvSortP` is indeed unreachable. But if you look at how `bvSortP` is defined, then it is:
```
bvSortP :: Parser Sort
bvSortP = mkSort <$> (bvSizeP "Size32" S32 <|> bvSizeP "Size64" S64)
  where
    bvSizeP ss s = do
      parens (reserved "BitVec" >> reserved ss)
      return s
```
So it recognises `(BitVec Sort32)` and `(BitVec Sort64)`. The strings also parse via the `(fAppTC <$> fTyConP <*> appArgsP)` option of this parser. So I think this is just redundant.

kosmikus: Actually, I think I'll comment out this branch, as I'm rather confident it is not having any effect anyway, and tests pass withou it.

kosmikus:
> Hi @kosmikus -- this is great, shall we merge? Was there a matching PR for `liquidhaskell` too?

@ranjitjhala I'm happy to merge this (and ucsd-progsys/liquidhaskell#1704) whenever you are ...

PR #447, opened 2020-10-01
Expose parts of the Horn parser

serras:
Hi! We are using `liquid-fixpoint` as part of a project, and it would be awesome if we could reuse parts of the parser for Horn clauses, and not only the top-level one.

Thanks in advance :)


ranjitjhala:
Totally! Please let me know if you get stuck etc. will be happy help!

PR #461, opened 2020-11-30
Expose Set_sng operation

serras:
It seems that the `Set_sng` did not work correctly. This PR exposes it in a similar way to the rest of the `Set` functions.


ranjitjhala:
Thank you so much!!!

PR #470, opened 2021-06-17
flatten and dedupe PAnd exprs

michaelborkowski:
In the BindEnvs used by the Solver, the refinement types can contain repetitions of the same conjuncts, which significantly increases the size of the larger Exprs and the resulting queries and SMT instances. This will speed up the solver.

This changes the pAnd smart constructor to flatten nested PAnd lists into a single level and then deduplicate.

### DIFF
--- a/src/Language/Fixpoint/Types/Refinements.hs
+++ b/src/Language/Fixpoint/Types/Refinements.hs
@@ -102,7 +102,7 @@ import           Data.Generics             (Data)
 import           Data.Typeable             (Typeable)
 import           Data.Hashable
 import           GHC.Generics              (Generic)
-import           Data.List                 (foldl', partition)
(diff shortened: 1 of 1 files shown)


michaelborkowski:
This commit is causing a few tests to diverge (proof/pleBoolA.fq, proof/pleBoolB.fq, proof/pleBoolC.fq), so I'm going to investigate that

nikivazou:
Thanks!!!! On what input flatten diverges????

michaelborkowski:
Fixed the original commit to make pAnd be order-preserving. The issue was that, if this is not the case, the evaluator in PLE can enter an infinite loop
```
PAnd [p, q] ~> PAnd [q, p] ~> PAnd [p, q] ~> ...
```

It looks like all tests pass now, so I'm marking this ready for consideration again.

PR #471, opened 2021-06-22
Dump the origin of constraints in .fq dumps

facundominguez:
When using `--savequery`, this PR includes the source origin of the constraints as a comment.

Before
```
$ cat .liquid/test.fq
...
// META wf : ()
...
```
After
```
$ cat .liquid/test.fq
...
// META wf : test.hs:13:1-4
...
```


ranjitjhala:
Thank you @facundominguez !!!

PR #472, opened 2021-06-25
Fix small glitch in 'stringPreamble'

serras:
Otherwise when `stringPreamble` is on, Z3 responds always:

```
invalid sorted variables: unknown sort 'string'
```


ranjitjhala:
Wow, thanks for catching/fixing that!!!

PR #474, opened 2021-07-16
Don't discard built artifacts before running tests in CI

facundominguez:


facundominguez:
Is calling `stack clean` before tests intentional?

It seems to be causing a rebuild of liquid-fixpoint in CI without an apparent reason.

ranjitjhala:
oops no :-(

facundominguez:
Well, that cut some 10 minutes of CI then :)

facundominguez:
Is this good to merge?

ranjitjhala:
thanks!!

PR #479, opened 2021-08-02
Allow to load back fq dumps into fixpoint

facundominguez:
There were some syntax discrepancies in the fq files produced by liquid-fixpoint, which prevented it from using these files as input.

* `match` clauses had parenthesis around the data constructor argument,
* `define` bodies weren't surrounded with braces,
* and `define` sorts weren't terminated with the sort of the body.

Moreover, liquid-fixpoint is only capable of parsing FInfos, while much of the processing happens on SInfos. This PR also implements a conversion of SInfos to FInfos.

### DIFF
--- a/src/Language/Fixpoint/Types/Constraints.hs
+++ b/src/Language/Fixpoint/Types/Constraints.hs
@@ -19,6 +19,7 @@ module Language.Fixpoint.Types.Constraints (
    -- * Top-level Queries
     FInfo, SInfo, GInfo (..), FInfoWithOpts(..)
   , convertFormat
+  , sinfoToFInfo
   , Solver
 
    -- * Serializing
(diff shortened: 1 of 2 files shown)


ranjitjhala:
Looks great! Should I merge?

facundominguez:
Yes, this is good to go. Maybe we could add at some point some test that `readFInfo` can parse the output of `writeFInfo`.

facundominguez:
@ranjitjhala,  ping :)

I have found that fq dumps produce:
```
match a C x y = (|| [x; y]) 
```
and the parser can't parse `||`. I can use instead
```
match a C x y = (if x then true else y)
```
but I'm not addressing that for now.

ranjitjhala:
Sorry I somehow thought I had merged already!

PR #487, opened 2021-08-31
Improve debugging experience

facundominguez:
This is a bundle of features that I implemented to help me understand liquid-fixpoint. I can split it if necessary.

Here's the summary:
* Implemented shortening of variable names in the prettified output. This is explained in the comments of function `proposeRenamings`.
* Various bits in .fq files were printed in undeterministic order, which made hard comparing different .fq files. This PR sorts these bits.
* ESym expressions were printed in a format that couldn't be parsed. I fixed this, so .fq files are loadable when they use ESym.
* Now bindings are omitted in environments of the prettified output if they start with an uppercase letter. I think this is a filter that accurately discards bindings corresponding to top-level definitions (which are uninteresting most of the time).
* The prettified output now prefixes unused bindings with an underscore "_". Previously, the binding name was replaced with an underscore, and this made hard correlating the prettified output with other dumps.

### DIFF
--- a/src/Language/Fixpoint/Solver/Prettify.hs
+++ b/src/Language/Fixpoint/Solver/Prettify.hs
@@ -15,6 +15,7 @@ import           Data.HashSet (HashSet)
 import qualified Data.HashSet as HashSet
 import           Data.List (intersperse, sortOn)
 import           Data.Maybe (fromMaybe)
+import           Data.Text (Text)
 import qualified Data.Text as Text
 import           Language.Fixpoint.Misc (ensurePath)
 import           Language.Fixpoint.Solver.EnvironmentReduction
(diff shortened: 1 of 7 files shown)


src/Language/Fixpoint/Solver/Prettify.hs
```diff
@@ -68,52 +90,158 @@ prettyConstraint aenv bindEnv c =
...
+proposeRenamings :: [Symbol] -> HashMap Symbol Symbol
+proposeRenamings = toSymMap . toPrefixSuffixMap
+
+-- | Indexes symbols by prefix and then by suffix. If the prefix
+-- equals the symbol, then the suffix is the empty symbol.
+--
+-- For instance,
+--
+-- > toPrefixSuffixMap ["a##b##c"] ! "a" ! "b" == ["a##b##c"]
+-- > toPrefixSuffixMap ["a##b"] ! "a" ! "" == ["a##b"]
```
facundominguez:
```suggestion
-- > toPrefixSuffixMap ["a##b##c"] ! "a" ! "c" == ["a##b##c"]
-- > toPrefixSuffixMap ["a"] ! "a" ! "" == ["a"]
--
-- In general,
--
-- > forall ss.
-- > Set.fromList ss == Set.fromList $ concat [ xs | m <- elems (toPrefixSuffixMap ss), xs <- elems m ]
-- 
-- > forall ss.
-- > and [ all (pfx `isPrefixOfSym`) xs && all (sfx `isSuffixOfSym`) xs
-- >     | (pfx, m) <- toList (toPrefixSuffixMap ss)
-- >     , (sfx, xs) <- toList m
-- >     ]
--
-- TODO: put the above in unit tests
```

nikivazou (approved):

PR #488, opened 2021-09-01
Small simplifications and documentation for PLE

facundominguez:
Besides explaining some functions and the purpose of the trie, this PR also simplifies the some functions.

The only commit with potential to break tests is the first one (7e8c466) Which I checked that doesn't break any test in LH (not when building with stack at least).

### DIFF
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -493,7 +493,7 @@ fastEval γ ctx e =
     go (ECoerc s t e)   = ECoerc s t  <$> go e
     go e@(EApp _ _)     = case splitEApp e of 
                            (f, es) -> do (f':es') <- mapM (fastEval γ ctx) (f:es)
-                                         evalApp γ ctx (eApps f' es) (f',es')
+                                         evalApp γ ctx f' es'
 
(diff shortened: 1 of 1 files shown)


src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -378,6 +421,9 @@ evalOne γ env ctx e = do
   env' <- execStateT (eval γ ctx [(e, PLE)]) (env { evFuel = icFuel ctx }) 
   return (evAccum env', evFuel env')
 
+-- | @notGuardedApps e@ yields all the subexpressions that are
+-- applications not under a lambda abstraction, type abstraction,
```
```suggestion
-- applications not under an if-then-else, lambda abstraction, type abstraction,
```

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -412,6 +458,16 @@ subsFromAssm (EEq lhs rhs) | (EVar v) <- unElab lhs
                            , anfPrefix `isPrefixOfSym` v = [(v, unElab rhs)]
 subsFromAssm _                                           = []
 
+-- | Unfolds expressions using rewrites and equations.
+--
+-- Also reduces if-then-else when the boolean condition or the negation can be
+-- proved valid.
```
```suggestion
-- proved valid. This is the actual implementation of guard-validation-before-unfolding
-- that is described in publications.
```

nikivazou (approved):

PR #490, opened 2021-09-02
Fix caching in ci

facundominguez:


facundominguez:
Seems to speed up from 22-23 to 15 minutes.

Issue #482, opened 2021-08-18
Optimize interaction with the SMT solver


facundominguez:
AmericanFlag.hs takes near 20 seconds to verify, while `z3 AmericanFlag.hs.smt2` takes only 2 seconds. It turns out most of the time is spent in `Language.Fixpoint.Smt.Interface.command` which is called 280000 times! Each invocation takes near 50 microseconds, which I conjecture is going to be difficult to speed up.

Maybe SMT solvers can be linked as libraries to cut on that, or maybe it would be possible to batch a few commands and see if just reducing the amount of exchanges with the SMT solver reduces the overall time.

Any thoughts?

facundominguez:
Found a couple of solver-agnostic apis:
https://github.com/makaimann/smt-switch
https://github.com/sosy-lab/java-smt (calling the jvm from Haskell is a bit slow though).

facundominguez:
There's also the chance that after studying the example we might find a way to verify it with less calls to the smt solver.

I tried [some batching](https://github.com/ucsd-progsys/liquid-fixpoint/compare/develop...facundominguez:fd/batching) that removed some `hFlush` calls on the pipe to the smt solver. This produced a speed up (70% of the original time), but apparently we need a different solution if we want to get rid of all of the overhead.

ranjitjhala:
Hi @facundominguez -- sorry for not replying, have been thinking this over. The data about `AmericanFlag` is very interesting -- I didn't realize we get a 50 ms hit _per_ call -- if my math is right, thats `280000 * 50 / 1000000 = 14s` (!) just spent in the interface? Does the 50ms include the serialization time? Or is it _only_ the invocation?

The batching sounds very promising! The catch in general is that there is a dependency but still, there could be significant gains as your numbers show -- does `AmericanFlag` drop to about `10s` with your edits? 
 
I have been wary of in-memory APIs because the `.smt` logs are so useful for debugging but this is a _lot_ of slowdown...

facundominguez:
> Does the 50ms include the serialization time?

Yes, it does include serialization.

> Does AmericanFlag drop to about 10s with your edits?

AmericanFlag overall goes from 18 seconds to 14 seconds. The time spent in the `command` function goes down to 10 seconds.

> I have been wary of in-memory APIs because the .smt logs are so useful for debugging but this is a lot of slowdown...

Perhaps LF could log just the same with a modest amount of work.

The main drawbacks of batching are:
[...]

PR #492, opened 2021-09-12
Bottleneck pruning in AmericanFlag

facundominguez:
Here go some optimizations driven by profiling liquid-fixpoint when fed with AmericanFlag.hs.bfq

These optimizations clean up the flamegraph in #482. The remaining bottlenecks seem to be in smt interactions via pipe, serialization, and HashMap/Set lookups.

### DIFF
--- a/src/Language/Fixpoint/Solver/EnvironmentReduction.hs
+++ b/src/Language/Fixpoint/Solver/EnvironmentReduction.hs
@@ -77,7 +77,7 @@ import           Language.Fixpoint.Types.Refinements
   , subst1
   )
 import           Language.Fixpoint.Types.Sorts (boolSort, sortSymbols)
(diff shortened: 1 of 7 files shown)


facundominguez:
By now, I discovered the scripts to build performance charts from logs. [Here](https://github.com/ucsd-progsys/liquid-fixpoint/files/7149582/perf1.svg.zip) are the results. Above the noise level, I only see some improvements of a few seconds in some benchmarks.

facundominguez:
... let me run one more performance check before merging this.

facundominguez:
Checked. All is good. Feel free to merge.

facundominguez:
Ping here :)

Do we need to do any more work to merge it?

I have some other optimizations to share, but I'd rather not grow this PR.

ranjitjhala:
Thanks Facundo!

PR #493, opened 2021-09-21
Optimize interactions with the SMT and serialization

facundominguez:
With these optimizations benchmarks come down another few seconds.


facundominguez:
Here is the comparison of benchmarks.
perf

I don't know why my attempt to optimize the hash function is failing in CI. But it is something that we could address later, certainly.

facundominguez:
Hello! Is this good to merge?

ranjitjhala:
Hi @facundominguez -- yes, I presume you tried it with LH as well to check all is OK there?

facundominguez:
Right. All tests pass for me in LH

curiousleo:
@facundominguez I'm curious: why did you revert https://github.com/ucsd-progsys/liquid-fixpoint/pull/493/commits/5befdfb1e80bbbbae369179380fbc90b39b2f768? I was just reading through the code and had the same idea :)

One reason I can think of would be that the `_symbolId` is perhaps not unique, i.e. there may be multiple `Symbol` instances with the same `_symbolId` but different `symbolRaw` or `symbolEncoded` -- is that why we need to hash on `symbolRaw`?

Edit: `Id` [is defined as](https://hackage.haskell.org/package/intern-0.9.4/docs/Data-Interned-Internal.html#t:Id) `type Id = Int`. If `_symbolId` _were_ unique then `FuelCount` could have `fcMap :: IntMap Int` instead of `fcMap :: M.HashMap Symbol Int` and we may not even need the `Hashable` instance at all. That could reduce work.

facundominguez:
IIRC, it did not do much of a difference in performance. Moreover 5befdfb is already reverting a1729ae, and I lacked the context of that change.

@gridaphobe, do you remember why the hash uses the symbol text rather than the integer identifier?

PR #495, opened 2021-10-04
Optimize serialization of bindings

### LINKED ISSUE #494: Serialize formulas only once
Currently, when we try to validate a constraint, we send to the SMT solver all the formulas in the environment and the right-hand-side of the constraint.

The problem is that many bindings in the environments of constraints are shared. So whenever multiple constraints are validated, all the shared bindings are serialized over and over, and serialization is relatively expensive. This is an important bottleneck in StitchLH.

I see two ways to reduce the duplicated effort. 

### 1. Cache the result of serializing a binding.

In this approach, we store together with each binding, the result of serializing it. Thus, whenever we need to validate a constraint that uses a binding, we send the cached bytestring to the SMT solver.

### 2. Load the bindings into the smt solver without asserting them.

How about we define a function for each binding?
```
define-fun f1 () Bool (long formula of binding 1)
define-fun f2 () Bool (long formula of binding 2)
```
Then when we need to validate a constraint we just say
```
assert (f1 && f2)
```
Instead of 
```
assert (long formula of binding 1 && long formula of binding 2)
```
I'm not very familiar with SMTLIB2, so maybe there is a better way to achieve the same effect.

I feel like trying (2) for a start. Any thoughts?

facundominguez:
This PR does a few clean ups and an optimization for serialization.

Before this PR, bindings of constraints were serialized for each constraint. If two constraints shared some bindings, these bindings would be serialized twice. This overhead turned out to be substantial for the Step.hs module of stitch-lh.

This PR reduces the overhead by defining functions for the bindings via smtlib2 as explained in #494. The benchmarks of LH do not improve much, but verification of Step.hs.bfq in LF goes from 17 seconds to 11 seconds. The overhead of serialization, initially at 40%, goes to 4%, and PLE takes most of the verification time now.

At the moment, bindings are serialized before the fixpoint calculation and after PLE. Two further improvements to do in subsequent PRs would be:
* Have PLE use the serialized bindings. Right now it ignores them.
* Have PLE not modify existing bindings as a result, but instead add new bindings with the new equations.

### DIFF
--- a/src/Language/Fixpoint/Solver/Solution.hs
+++ b/src/Language/Fixpoint/Solver/Solution.hs
@@ -268,10 +268,16 @@ okInst env v t eq = isNothing tc
 -- | Predicate corresponding to LHS of constraint in current solution
 --------------------------------------------------------------------------------
 {-# SCC lhsPred #-}
-lhsPred :: (F.Loc a) => F.BindEnv -> Sol.Solution -> F.SimpC a -> F.Expr
-lhsPred be s c = F.notracepp _msg $ fst $ apply g s bs
+lhsPred
+  :: (F.Loc a)
+  => F.IBindEnv
+  -> F.BindEnv
+  -> Sol.Solution
+  -> F.SimpC a
+  -> F.Expr
+lhsPred bindingsInSmt be s c = F.notracepp _msg $ fst $ apply g s bs
   where
-    g          = CEnv ci be bs (F.srcSpan c)
+    g          = CEnv ci be bs (F.srcSpan c) bindingsInSmt
     bs         = F.senv c
     ci         = sid c
     _msg       = "LhsPred for id = " ++ show (sid c) ++ " with SOLUTION = " ++ F.showpp s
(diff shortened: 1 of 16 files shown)


facundominguez:
Here are the unimpressive benchmark results of LH
perf

Here's the profile of Step.hs.bfq verification with this optimization: [fixpoint.after.svg.zip](https://github.com/ucsd-progsys/liquid-fixpoint/files/7275039/fixpoint.after.svg.zip)
It can be compared with the profile shared in #494.

src/Language/Fixpoint/Solver/Solution.hs
```diff
@@ -292,7 +302,10 @@ type ExprInfo    = (F.Expr, KInfo)
 apply :: CombinedEnv -> Sol.Sol a Sol.QBind -> F.IBindEnv -> ExprInfo
 apply g s bs      = (F.conj (pks:ps), kI)   -- see [NOTE: pAnd-SLOW]
   where
-    (pks, kI)     = applyKVars g s ks  
+    -- Clear the "known" bindings for applyKVars, since it depends on
+    -- using the fully expanded representation of the predicates to bind their
+    -- variables with quantifiers.
```
facundominguez, line 307: This is a shortcoming that wouldn't occur should we implement the caching described in approach (1) of #494. That said, the overhead seems pretty small currently to go for it.

ranjitjhala:
Sweet! Ready to merge?

ranjitjhala (approved):
This is wonderful, thank you so much!

facundominguez:
Yes, this is good to go. I tested in LH that no test breaks.

PR #497, opened 2021-10-14
Updates to Liquid Fixpoint for REST

zgrannan:
Improve rewriting. Related: https://github.com/ucsd-progsys/liquidhaskell/pull/1882

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
           name: Dependencies
           command: |
             stack setup
-            stack build liquid-fixpoint --only-dependencies --test --no-run-tests
+            stack build -j2 liquid-fixpoint --only-dependencies --test --no-run-tests
       - save_cache:
           key: stack-{{ checksum "stack.yaml" }}
           paths:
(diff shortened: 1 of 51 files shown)


facundominguez:
The conflicts in PLE.hs and Solve.hs are on changes that I introduced. I can help if any of them look confusing.

nikivazou:
@zgrannan this is not ready to merge, correct?

zgrannan:
@nikivazou It's almost ready; the last part is to upload the REST library to Hackage. Currently the dependency is to my Github repo, which might not be ideal (since it won't work with Cabal I assume).

However, I am still waiting for hackage to give me upload permission, as apparently they need to manually approve that part.

nikivazou:
Perfect! We can wait until you get the permission. 
So, if there are no active rewrites nothing should be changed, correct? 
(@facundominguez do you have PLE performance regression tests that we could check that?)

facundominguez:
In general, I look at the tests in liquidhaskell for performance regressions. [Here](https://github.com/ucsd-progsys/liquidhaskell#how-to-create-performance-comparison-charts) there are some notes on how to compare executions with and without a change.

I also found costs in stitch-lh that are not accounted for in the current test suite (this far, only related to SMTLIB2 overheads). Checking this would amount to observing how long it takes to build stitch-lh with and without the change.

nikivazou:
Got it! Thanks a lot @facundominguez!

nikivazou (approved):

PR #499, opened 2021-10-27
Minor optimizations for PLE

facundominguez:
All LH tests continue to pass.
The first and the third commit give me a 10% speed up in stitch-lh.
The other commits are just cleanups without an apparent change in performance.


facundominguez:
I kind of think I know what I'm doing when messing with PLE. But wouldn't harm to have someone look over my shoulder :)

facundominguez:
I solved conflicts and rerun LH tests. All of them pass.

nikivazou:
Thanks a lot!

PR #502, opened 2021-11-02
Interpreter ready to merge into fixpoint

michaelborkowski:
Once this is merged into fixpoint I can do a PR for the liquidhaskell repo


src/Language/Fixpoint/Solver/Interpreter.hs
```diff
@@ -0,0 +1,887 @@
...
+      unElabRefts (sym, sr) = let r = sr_reft sr in
+                                (,) (EVar sym) <$> (unElab <$> isSingletonE (reftBind r) (reftPred r))-}
+
+evalCandsLoop :: ConstMap -> Config -> ICtx -> Knowledge -> EvalEnv -> InstRes -> IO ICtx 
+evalCandsLoop ie _ ictx0 γ env res = go ictx0 
+  where
+    withRewrites exprs =
+      let
+        rws = [rewrite e rw | rw <- snd <$> M.toList (knSims γ)
+                            ,  e <- S.toList (snd `S.map` exprs)]
```
facundominguez, line 133: The RHSs were superfluous here in the previous PLE. All tests would pass without them, and they are still added as candidates in `newcands` below. Are they needed here for some particular reason?

src/Language/Fixpoint/Solver/Interpreter.hs
```diff
@@ -0,0 +1,887 @@
...
+    isSingletonE v (PAnd cs) =
+      msum $ map (isSingletonE v) cs
+    isSingletonE _ _ =
+      Nothing
+
+    isSingEq v e0 e1 = do
+      guard $ EVar v == dropECst e0 && not (S.member v $ exprSymbolsSet e1)
+      Just e1
+
+    isEq r = r == Eq || r == Ueq-}
```
facundominguez: Perhaps these comments ended up here accidentally. Should they be removed?

src/Language/Fixpoint/Solver/Interpreter.hs
```diff
@@ -0,0 +1,887 @@
...
+
+--------------------------------------------------------------------------------
+
+eqArgNames :: Equation -> [Symbol]
+eqArgNames = map fst . eqArgs
+
+interpret' :: ConstMap -> Knowledge -> ICtx -> SEnv Sort -> InstRes -> Expr -> Expr
+interpret' ie γ ctx env res e = mytracepp ("Interpreting " ++ show e) $ interpret ie γ ctx env res e
+
+interpret :: ConstMap -> Knowledge -> ICtx -> SEnv Sort -> InstRes -> Expr -> Expr
```
facundominguez: Some documentation here explaining how the expression is transformed would help. In particular, I'm curious about the role of the `InstRes` parameter, but also how much it leaves unevaluated.

src/Language/Fixpoint/Solver/Interpreter.hs
```diff
@@ -0,0 +1,887 @@
...
+{-    where
+      anfBEnv = filterBindEnv anfFilter ieBEnv 
+      anfFilter = (\i sym r -> (anfPrefix `isPrefixOfSym` sym) || (vvName `isPrefixOfSym` sym))
+
+      anfEnv = M.fromList ( Mb.mapMaybe unElabRefts (M.elems (beBinds anfBEnv)) )
+
+      unElabRefts (sym, sr) = let r = sr_reft sr in
+                                (,) (EVar sym) <$> (unElab <$> isSingletonE (reftBind r) (reftPred r))-}
+
+evalCandsLoop :: ConstMap -> Config -> ICtx -> Knowledge -> EvalEnv -> InstRes -> IO ICtx
```
facundominguez: Could we have documented here what the `res` parameter is for? This seems to be a difference with respect to the previous PLE.

src/Language/Fixpoint/Solver/Interpreter.hs
```diff
@@ -0,0 +1,887 @@
...
+mytracepp :: (PPrint a) => String -> a -> a
+mytracepp = notracepp 
+
+mytrace :: String -> a -> a
+mytrace = {-trace-} flip const
+
+--------------------------------------------------------------------------------
+-- | Strengthen Constraint Environments via PLE 
+--------------------------------------------------------------------------------
+instInterpreter :: (Loc a) => Config -> SInfo a -> Maybe [SubcId] -> IO (SInfo a)
```
facundominguez, line 53: Either here or in the module documentation, it would help offering context on how this PLE version differs from others.

Just as an example of puzzlement, perhaps this implementation tries to discover equations without any help at all from the SMT solver. This could be made a tad more explicit.

facundominguez (commented):
Hello! I'm sharing some comments in case they are useful.

I will have to study the code more to understand it properly, but don't wait for me to merge.

src/Language/Fixpoint/Solver/Solve.hs
```diff
@@ -110,16 +125,40 @@ solve_ cfg fi s0 ks wkl = do
...
+{-
+printUnsolved :: (NFData a, F.Fixpoint a, F.Loc a) => Config -> F.SInfo a -> [F.SubcId] -> Doc
+printUnsolved cfg fi bads = csDoc $+$ text "\n"
+  where
+    cs            = [ (i, c) | (i, c) <- M.toList (F.cm fi)
+                      ,  i `L.elem` bads ]
+--    cm'           = M.filterWithKey (\id v -> L.elem id bads) . F.cm
+    csDoc         = vcat     . map (toFix . snd) $ {-hashMapToAscList $-} cs
+-}
+
```
facundominguez, line 149: More comments that seem accidental.

ranjitjhala:
These are excellent comments - Michael can you address/fix please ? Thanks!

src/Language/Fixpoint/Solver/Interpreter.hs
```diff
@@ -0,0 +1,887 @@
...
+  , knAllDCs                   = S.fromList $ (val . dcName) <$> concatMap ddCtors (ddecls si)
+  , knSels                     = M.fromList . Mb.catMaybes $ map makeSel  sims 
+  , knConsts                   = M.fromList . Mb.catMaybes $ map makeCons sims 
+  , knAutoRWs                  = aenvAutoRW aenv
+  , knRWTerminationOpts        = if (rwTerminationCheck cfg)
+                                   then RWTerminationCheckEnabled 
+                                   else RWTerminationCheckDisabled
+  } 
+  where 
+    sims = aenvSimpl aenv ++ concatMap reWriteDDecl (ddecls si) -- store as list too?
```
facundominguez:
```suggestion
    sims = aenvSimpl aenv
```
This is another change that went to the previous PLE.

src/Language/Fixpoint/Types/Config.hs
```diff
@@ -183,6 +184,7 @@ defConfig = Config {
   , nonLinCuts               = False &= help "Treat non-linear kvars as cuts"
   , noslice                  = False &= help "Disable non-concrete KVar slicing"
   , rewriteAxioms            = False &= help "allow axiom instantiation via rewriting"
+  , useInterpreter           = False &= help "Use the interpreter to assist PLE"
```
facundominguez: If I'm reading this correctly, defaulting to `False` means that no test in CI rehearses this algorithm. Shouldn't it be enabled for some of the tests? The whole testsuite is fast enough that maybe it could be executed a second time with this flag enabled.

ranjitjhala:
I agree we should rerun with the flag on

Michael can you add that please?

src/Language/Fixpoint/Solver/Interpreter.hs
```diff
@@ -0,0 +1,887 @@
...
+equalitiesPred eqs = [ EEq e1 e2 | (e1, e2) <- S.toList eqs, e1 /= e2 ] 
+
+updCtxRes :: InstRes -> Maybe BindId -> ICtx -> (ICtx, InstRes) 
+updCtxRes res iMb ctx = (ctx, res')
+  where 
+    res' = updRes res iMb (pAnd $ equalitiesPred $ icEquals ctx)
+
+
+updRes :: InstRes -> Maybe BindId -> Expr -> InstRes
+updRes res (Just i) e = M.insert i e res
```
facundominguez: This algorithm seems to depend on the tree-like invariant too. Here it could check it as done in https://github.com/ucsd-progsys/liquid-fixpoint/pull/498/commits/a0f2dc95ed9b3c7606ceea5a17db97ce40c08048

facundominguez: Perhaps it is worth factoring the common functions so the code isn't duplicated for the sake of these changes. It may not necessarily be the most pragmatic if there are subtle differences between the copies.

src/Language/Fixpoint/Solver/Interpreter.hs
michaelborkowski: Yes, one big part of the interpreter idea was to avoid any interaction with the SMT solver, which also avoids the expensive I/O associated with the communication. I'll add to the documentation

src/Language/Fixpoint/Solver/Interpreter.hs
michaelborkowski: The InstRes object is being passed around, but it's no longer being used for anything in `evalsCandsLoop` or anything it calls. I'm removing it here. Thanks for pointing that out.

src/Language/Fixpoint/Solver/Interpreter.hs
```diff
@@ -0,0 +1,843 @@
...
+             su'  = makeSu ess 
+    makeSu exprs  = [(x,c) | (EEq (EVar x) c) <- exprs 
+                           , isConstant (knDCs γ) c 
+                           , EVar x /= c ]
+
+makeCandidates :: Knowledge -> ICtx -> Expr -> [Expr]
+makeCandidates γ ctx expr 
+  = mytracepp ("\n" ++ show (length cands) ++ " New Candidates") cands
+  where 
+    cands = filter (\e -> isRedex γ e && not (e `S.member` icSolved ctx)) (notGuardedApps expr)
```
facundominguez: This code has been fixed in the smt-based ple in https://github.com/ucsd-progsys/liquid-fixpoint/pull/504.

src/Language/Fixpoint/Solver/Solve.hs
```diff
@@ -110,12 +125,26 @@ solve_ cfg fi s0 ks wkl = do
     res0     <- {- SCC "sol-result" #-} result bindingsInSmt cfg wkl s3
     return (s3, res0)
   res <- case resStatus res0 of
-    Unsafe _ bads | not (noLazyPLE cfg) && rewriteAxioms cfg -> do
+    Unsafe _ bads | not (noLazyPLE cfg) && rewriteAxioms cfg && useInterpreter cfg -> do
```
facundominguez, line 128: Is it necessary to choose between the two PLE implementations? I would have expected a first pass with the interpreter, and then a second pass on the remaining unsolved constraints with the smt-based PLE.

src/Language/Fixpoint/Solver/Interpreter.hs
```diff
@@ -0,0 +1,856 @@
...
+interpret _  _ _   _   e@(PGrad _ _ _ _) = e
+interpret ie γ ctx env (ECoerc s t e)    = let e' = interpret' ie γ ctx env e in
+                                             if s == t then e' else (ECoerc s t e')
+
+        
+--------------------------------------------------------------------------------
+-- | Knowledge (SMT Interaction)
+--------------------------------------------------------------------------------
+data Knowledge = KN 
+  { knSims              :: M.HashMap (Symbol, Symbol) Rewrite  -- ^ Rewrite rules came from match and data type definitions
```
facundominguez, line 555: `knSims` was changed in the smt-based PLE to `Map Symbol Rewrite` which allows faster lookups there.

At this point I'm not sure if it is worth insisting on porting every change. Maybe we should just agree to address the differences whenever we factorize the common parts of both PLEs.

src/Language/Fixpoint/Solver/Interpreter.hs
facundominguez: > This algorithm seems to depend on the tree-like invariant too. Here it could check it as done in a0f2dc9

This has been addressed.

michaelborkowski:
This is all ready to go. The interpreter is on by default now. Attached is an image of the benchmarks 
(The benchmarks are in a modified format: no --diff, using --no-check-imports and deleting the .hi/.o files beforehand so that the tests aren't skipped -- will make a separate issue in LH about making the benchmarks more reproducible)
develop-michael-fix-1

ranjitjhala:
Sweet don’t forget the matching PR for LH

PR #507, opened 2021-11-09
Remove trailing whitespace in Haskell files

curiousleo:
Here's what I ran to make this change:

    git ls-tree -z -r HEAD --name-only |
        grep -z '\.hs$' |
        xargs -0 -I{} sed -E -i 's/\s+$//g' {}

The motivation is that I was playing with applying `hlint` suggestions locally and found many changes that looked like nothing had changed at all. As it turns out, one of the things `hlint` fixes is trailing whitespace!

I think that automated refactors can be useful and may follow this up with a few refactors suggested by code analysis tools.

This PR prepares the ground by _only_ tackling trailing whitespace: there are no code or comment changes in this PR whatsoever. This makes actual refactors that need to be reviewed more carefully (which may come later) much easier to review.

The downside is that this is a change that will show up e.g. in `git blame`, to some degree masking the actual origin of a particular line of code. So it's a judgment call as to whether this change is worth merging.


ranjitjhala (approved):

ranjitjhala:
Haha thank you! I find those trailing spaces most bothersome too :-)

PR #508, opened 2021-11-10
Interpreter only on LHS

nikivazou:


facundominguez:
@nikivazou is this something worth pursuing to completion now the the interpreter has been merged?

I lack some context to be sure, but two things seem to be happening in this PR that look worth finishing: Firstly, code duplication in PLE and the interpreter is factored out. Secondly, there is an attempt to simplify the interpreter by discarding the Trie that PLE uses, and instead going with a smaller procedure (maybe you can give an overview of the differences).

nikivazou:
Exactly! There is the refactoring and code is much simpler. But this interpreter is not strong enough for @michaelborkowski 's case study to pass. 
So, my suggestion is to leave it hugging until I find time to improve it to handle the case study...

facundominguez:
Hello! Is this PR still helpful to keep open?

PR #509, opened 2021-11-10
Fix Nix + Stack setup

### LINKED ISSUE #506: `nix-build` and `nix-shell` error: "called without required argument"
Hi, the Nix setup isn't working for me on b02601f49f01578bd5158a5d29bf0e1c685b7f89.

```console
$ uname -a
Linux phoebe 5.10.76 #1-NixOS SMP Wed Oct 27 07:56:57 UTC 2021 x86_64 GNU/Linux
```

```console
$ nix-shell
error: anonymous function at /nix/store/nj5pzc098vcxifbla0nd9wxpyg03vlj8-cabal2nix-liquid-fixpoint/default.nix:1:1 called without required argument 'rest-rewrite', at /nix/store/2gvpjwk27dng6c6vw3q3dpfdbza91wmn-source/pkgs/development/haskell-modules/make-package-set.nix:87:27

$ nix-build .
error: anonymous function at /nix/store/nj5pzc098vcxifbla0nd9wxpyg03vlj8-cabal2nix-liquid-fixpoint/default.nix:1:1 called without required argument 'rest-rewrite', at /nix/store/2gvpjwk27dng6c6vw3q3dpfdbza91wmn-source/pkgs/development/haskell-modules/make-package-set.nix:87:27

$ nix-build . -A haskellPackages.liquid-fixpoint
error: anonymous function at /nix/store/nj5pzc098vcxifbla0nd9wxpyg03vlj8-cabal2nix-liquid-fixpoint/default.nix:1:1 called without required argument 'rest-rewrite', at /nix/store/2gvpjwk27dng6c6vw3q3dpfdbza91wmn-source/pkgs/development/haskell-modules/make-package-set.nix:87:27
```

curiousleo:
Fixes #506. Mostly this is a translation of the setup used in https://github.com/ucsd-progsys/liquidhaskell.

# Things that work

I have checked that the following work (tested on NixOS):

```
$ nix-shell --pure --run 'stack build'
$ nix-shell --pure --run 'stack test'
$ nix-shell --pure --run 'stack exec fixpoint -- tests/pos/adt.fq'
$ nix-shell --pure --run 'stack install'
```

The commands run via `nix-shell --pure --run` of course also run without `nix-shell`, assuming that the relevant tools are installed.

# Documentation

See the rendered preview of the updated README [here](https://github.com/curiousleo/liquid-fixpoint/tree/update-nix#how-to-build-and-install).

# Differences to the previous Nix setup

- Pinned an updated version of Nixpkgs.
- Updated from `ghc8101` to `ghc8107` out of convenience: the new Nixpkgs pin does not contain `ghc8101`. This brings the Nix setup in line with `stack.yaml`, which already uses `compiler: ghc-8.10.7`.
- `default.nix` previously overrode some dependencies. That no longer seems necessary, but I may have missed something here.
- Other things? The Nix setup is not documented, so I'm not sure what other commands are meant to be available or how else the Nix shell is meant to be used.


liquid-fixpoint.cabal
```diff
@@ -130,7 +130,7 @@ library
                   , mtl
                   , parallel
                   , parser-combinators
-                  , megaparsec           >= 7.0.0 && < 9
+                  , megaparsec           >= 7.0.0 && < 10
```
curiousleo, line 133: Compiles just fine with `megaparsec` version 9.

shell.nix
```diff
@@ -1 +1,12 @@
...
+{pkgs ? import ./nixpkgs.nix {}}:
+
+with pkgs;
+
+mkShell {
+  LANG="C.UTF-8";
+
+  buildInputs = [
+    ghc
+    ghcid
```
facundominguez: No z3 in the inputs?

Have you tested the setup with `nix-shell --pure`?

I'd leave `ghcid` out as it is non-essential to building. But I don't feel strong about it.

Also, I'd prefer to be explicit with the `ghc` compiler version and use instead `haskell.compiler.ghc8107` as input, just so it is easier to guess by looking at the configuration.

facundominguez (commented):
Thanks @curiousleo! It looks so much simpler.

> The Nix setup is not documented,

Good point. Perhaps a short section in the readme could say how to build with nix.

shell.nix
curiousleo: > Have you tested the setup with `nix-shell --pure`?

As a matter of personal preference, I use project-specific Nix shells as "additions" to my normal tools, which include nonstandard things like `rg`, `fd`, `zoxide`, `fzf`, etc. So I don't myself use `nix-shell --pure`.

The more relevant question is: what would you like me to test inside `nix-shell --pure`?

I have listed the commands I tested in the description. The declarations in `shell-stack.nix` bring `z3` into scope for `stack exec fixpoint -- --solver=z3` to succeed.

What does not work in the current setup is:

```
$ stack install
$ fixpoint tests/pos/adt.fq
Liquid-Fixpoint Copyright 2013-21 Regents of the University of California.
All Rights Reserved.

Working   0% [.................................................................]
fixpoint: fd:13: hGetLine: end of file
```

This is fixed by adding `z3` to `buildInputs` in `shell.nix`.

Is `stack install && fixpoint ...` something that should be supported in addition to `stack exec fixpoint -- ...`? It sounds to me like you're suggesting it should, so I'll add `z3` and extend the description.

---

> I'd leave `ghcid` out as it is non-essential to building. But I don't feel strong about it.

Done.

---

> Also, I'd prefer to be explicit with the `ghc` compiler version and use instead `haskell.compiler.ghc8107` as input, just so it is easier to guess by looking at the configuration.

Makes sense, done.

curiousleo: (Note that I have updated the description to list the commands that work now.)

curiousleo:
> Perhaps a short section in the readme could say how to build with nix.

This seems important: the setup in this PR allows you to build the project with `stack`, which now uses Nix internally. I don't this PR actually provides something that could be built with `nix-build` (which is what I think of when I read "build with nix").

In this sense, the README is perhaps a bit short on details but generally correct: `stack install` stays the preferred way to build and install `fixpoint` locally.

- Would you like to be able to build the project (specifically the `fixpoint` executable, I guess) with `nix-build`?
- Assuming that we're happy to build with `stack`, what additional documentation would you like to see in the README?

facundominguez:
> The more relevant question is: what would you like me to test inside nix-shell --pure?
> Is `stack install && fixpoint` ... something that should be supported in addition to `stack exec fixpoint -- ...`?

hm, fair questions. I hadn't noticed `z3` in `shell-stack.nix`. I have called z3 directly when working in the liquidhaskell repo, but not as part of a regular build. Also, I needed to test `liquidhaskell` with `cabal-install`, in which case I also needed `z3` in the shell.

I think in these questions it is reasonable to defer to your judgement, or if other users/devs of `liquid-fixpoint` have opinions they can chime in.

> As a matter of personal preference, I use project-specific Nix shells as "additions" to my normal tools

I see. So in a project with multiple developers we would end up installing the union of all the tools that any developer might care about?

> Would you like to be able to build the project (specifically the fixpoint executable, I guess) with nix-build?

Not really. When I build I use
```
$ nix-shell --pure --run "stack build"
```
which would offer the most reproducible build. This even needs stack to be listed in the inputs of the shell.

> Assuming that we're happy to build with stack, what additional documentation would you like to see in the README?

In essence:
```
# Build with
$ nix-shell --pure --run "stack build"
# Test with
$ nix-shell --pure --run "stack test"
# Run with
$ nix-shell --pure --run "stack exec fixpoint -- tests/pos/adt.fq"
```
The user is free to drop the `--pure` or tweak the shell at the risk of introducing differences in behavior.

tests/test.hs
```diff
@@ -373,7 +373,7 @@ loggingTestReporter = TestReporter [] $ \opts tree -> Just $ \smap -> do
                        "test, time(s), result"]
 
 
-    let dir = "tests" </> "logs" </> host ++ "-" ++ time
+    -- let dir = "tests" </> "logs" </> host ++ "-" ++ time
```
facundominguez:
```suggestion
```

tests/test.hs
```diff
@@ -358,7 +358,7 @@ loggingTestReporter = TestReporter [] $ \opts tree -> Just $ \smap -> do
 
   return $ \_elapsedTime -> do
     -- get some semblance of a hostname
-    host <- takeWhile (/='.') . takeWhile (not . isSpace) <$> readProcess "hostname" [] []
+    -- host <- takeWhile (/='.') . takeWhile (not . isSpace) <$> readProcess "hostname" [] []
```
facundominguez:
```suggestion
```
Good catch.

facundominguez (approved):
Looks very good to me!

facundominguez:
I'll merge it tomorrow if nobody has further comments.

curiousleo:
> > The more relevant question is: what would you like me to test inside nix-shell --pure?
> > Is `stack install && fixpoint` ... something that should be supported in addition to `stack exec fixpoint -- ...`?
> 
> hm, fair questions. I hadn't noticed `z3` in `shell-stack.nix`. I have called z3 directly when working in the liquidhaskell repo, but not as part of a regular build. Also, I needed to test `liquidhaskell` with `cabal-install`, in which case I also needed `z3` in the shell.
> 
> I think in these questions it is reasonable to defer to your judgement, or if other users/devs of `liquid-fixpoint` have opinions they can chime in.

For now I've gone with the minimal `shell.nix` that makes the `nix-shell --pure --run 'stack ...'` commands work. This includes GHC, stack and Nix but does not include Z3.

> > As a matter of personal preference, I use project-specific Nix shells as "additions" to my normal tools
> 
> I see. So in a project with multiple developers we would end up installing the union of all the tools that any developer might care about?

To clarify: I think that `--run` should pretty much always be used in combination with `--pure`.

I was talking about setting up a project-specific shell environment that I would switch into with `nix-shell<Enter>` or via direnv. When I do that, I still want to have `vim`, `rg` etc. available.

To your point -- yes, if the `shell.nix` is used in this way, developers need to agree on what they consider useful project specific tooling. I have not found this to be very contentious in the past. If a project requires frequent use of Terraform or Kubernetes then it makes sense to add `terraform` and `kubernetes` to the environment ...

In this sense, I think that adding Z3 to `shell.nix` is not controversial. I haven't done it yet because "everything needed to run `nix-shell --pure --run 'stack ...'`" is such a neat, succinct definition that I've run with it in this PR. I would be happy to see a follow-up that adds Z3 and perhaps other tools people use for development to `shell.nix`.

> In essence:
> 
> ```
> # Build with
> $ nix-shell --pure --run "stack build"
> # Test with
> $ nix-shell --pure --run "stack test"
> # Run with
> $ nix-shell --pure --run "stack exec fixpoint -- tests/pos/adt.fq"
> ```

Excellent. I have added this to the README.

tests/test.hs
```diff
@@ -373,7 +372,6 @@ loggingTestReporter = TestReporter [] $ \opts tree -> Just $ \smap -> do
                        "test, time(s), result"]
 
 
-    let dir = "tests" </> "logs" </> host ++ "-" ++ time
```
curiousleo, line 376: The only use of `dir`, 5 lines further down, is commented out. This definition of `dir`, in turn, is the only place where `host` is used. That allowed me to comment out the definition of `host`, which is the only place that needed the `hostname` command, which allowed me to leave `hostname` out of `shell.nix`.

curiousleo: Now removed instead of commented out.

curiousleo:
@ranjitjhala, any thoughts on this proposed change? This is the most minimal functioning Nix/Stack setup that I could think of.

curiousleo:
Good to go, I think!

PR #511, opened 2021-11-10
Apply hlint suggestions to PLE.hs

curiousleo:
Applies some of `hlint`'s suggestions.

My aim is to simplify code and make it more readable. "Readable" is subjective, so feel free to disagree with the suggestions here! I'll be happy to oblige.

I have made the painful experience that even innocent-looking changes like this can affect performance in unforeseen ways. So I would like to benchmark these changes against `develop` before merging. That is why I have marked this as a draft for now.

### DIFF
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -90,7 +90,7 @@ instantiate cfg fi' subcIds = do
   where
     withRESTSolver :: (Maybe SolverHandle -> IO a) -> IO a
     withRESTSolver f | null (concat $ M.elems $ aenvAutoRW aEnv) = f Nothing
-    withRESTSolver f | otherwise = withZ3 (\z3 -> f (Just z3))
+    withRESTSolver f  = withZ3 (f . Just)
 
     file   = srcFile cfg ++ ".evals"
     sEnv   = symbolEnv cfg fi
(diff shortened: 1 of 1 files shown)


src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -470,20 +470,17 @@ getAutoRws γ ctx =
...
   where
     oc :: AbstractOC (OCType Op) Expr IO
     oc = ordConstraints (Mb.fromJust $ restSolver env)
 
     rp = RP oc [(e, PLE)] constraints
-    constraints = foldl go (OC.top oc) []
-      where
-        go c (t, u) = refine oc c t u
-
+    constraints = OC.top oc
```
curiousleo, line 483: This is probably the most interesting change in this PR.

`foldl go (OC.top oc) []`, which reduces to `OC.top oc`, appears to be a leftover from this change: https://github.com/ucsd-progsys/liquid-fixpoint/commit/78e875e099c59a9d5e9f138bebba2ec157cd3c88#diff-138cb912182d4ea38121c4b70aef046e31f048a879f00bdfa2a03cfd0ac5f5e5L468-R477

facundominguez (approved):
All changes look good.

The way to test for performance regressions would be to run the benchmarks of liquidhaskell with these changes.

Note that between running benchmarks it is necessary to remove a bunch of intermediate files.

```
$ find benchmarks/ \( -name "*.hi" -o -name "*.o" -o -name .liquid -o -name "*hs_outputdir" -o -name "*hs_odir" \) -exec rm -r {} ";"
$ find tests/ \( -name "*.hi" -o -name "*.o" -o -name .liquid -o -name "*hs_outputdir" -o -name "*hs_odir" \) -exec rm -r {} ";"
```
Actually, this should be mentioned in the README of liquidhaskell but it isn't.

curiousleo:
Before = 5dcfd3b1e2c752e6f0f727f65e837155d6fc2b9f
After = 095bf7e05d6044c5db835e2fb0bcf1c2540ea527

It looks to me like the difference is well within the measurement error, which is what I would hope for in a small refactoring PR like this. Still, I'm glad I checked.

perf

curiousleo:
Good to go, I think.

ranjitjhala:
Thanks you!!!

PR #512, opened 2021-11-12
Fix links in README.md

curiousleo:
- Z3 and CVC4 appear to have moved their main presence to GitHub.
- The `ple_list03.smt2` file doesn't exist, so I removed the link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@ Requirements
 
 In addition to the .cabal dependencies you require an SMTLIB2 compatible solver binary:
 
-- [Z3](http://z3.codeplex.com)
-- [CVC4](http://cvc4.cs.nyu.edu)
+- [Z3](https://github.com/Z3Prover/z3)
+- [CVC4](https://cvc4.github.io/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,7 +71,6 @@ See the examples in `tests/horn/{pos, neg}` eg
 
 - [sum](tests/horn/pos/ple_sum.smt2)
 - [list00](tests/horn/pos/ple_list00.smt2)
-- [list03](tests/horn/pos/ple_list03.smt2)
```
facundominguez, line 74: This could be `tests/horn/neg/ple_list03.smt2` instead.

facundominguez (commented):
Thanks Leo! The old links are broken for me as well, and the new ones work.

I'm not sure about removing the example. Perhaps the path is wrong.

README.md
curiousleo: Good call! I've added this link back in, adjusted to point to `neg` rather than `pos`.

ranjitjhala:
Thanks for fixing the links!

facundominguez (approved):

PR #517, opened 2021-11-23
Formatting and documentation

curiousleo:
I left a couple of questions in comments in `normalizeBody`.

### DIFF
--- a/src/Language/Fixpoint/Defunctionalize.hs
+++ b/src/Language/Fixpoint/Defunctionalize.hs
@@ -112,7 +112,7 @@ instance Defunc (SimpC a) where
   defunc sc = do crhs' <- defunc $ _crhs sc
                  return $ sc {_crhs = crhs'}
 
-instance Defunc (WfC a)   where
+instance Defunc (WfC a) where
   defunc wf@(WfC {}) = do
     let (x, t, k) = wrft wf
     t' <- defunc t
@@ -141,7 +141,7 @@ instance Defunc Expr where
 instance Defunc a => Defunc (SEnv a) where
   defunc = mapMSEnv defunc
 
-instance Defunc BindEnv   where
+instance Defunc BindEnv where
   defunc bs = do dfbs <- gets dfBEnv
                  let f (i, xs) = if i `memberIBindEnv` dfbs
                                        then  (i,) <$> defunc xs
@@ -152,7 +152,7 @@ instance Defunc BindEnv   where
     -- the bind does not appear in any contraint,
     -- thus unique binders does not perform properly
     -- The sort should be defunc, to ensure same sort on double binders
-    matchSort (x, RR s r) = ((x,) . (`RR` r)) <$> defunc s
+    matchSort (x, RR s r) = (x,) . (`RR` r) <$> defunc s
 
 -- Sort defunctionalization [should be done by elaboration]
 instance Defunc Sort where
@@ -164,7 +164,7 @@ instance Defunc a => Defunc [a] where
 instance (Defunc a, Eq k, Hashable k) => Defunc (M.HashMap k a) where
   defunc m = M.fromList <$> mapM (secondM defunc) (M.toList m)
 
-type DF    = State DFST
+type DF = State DFST
 
 data DFST = DFST
   { dfFresh :: !Int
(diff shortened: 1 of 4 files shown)


src/Language/Fixpoint/Defunctionalize.hs
```diff
@@ -203,8 +203,3 @@ freshSym t = do
   let x = intSymbol "lambda_fun_" n
   modify $ \s -> s {dfFresh = n + 1, dfBinds = insertSEnv x t (dfBinds s)}
   return x
-
-
--- | getLams and getRedexes return the (previously seen) lambdas and redexes,
---   after "closing" them by quantifying out free vars corresponding to the
---   fresh binders in `dfBinds`.
```
curiousleo, line 210: `getLams` and `getRedexes` no longer exist.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -1154,42 +1161,36 @@ instance Normalizable Equation where
...
-      | f `elem` syms e
-      = go' e
-    go e
-      = e
-
-    go' (PAnd [PImp c e1,PImp (PNot c') e2])
-      | c == c' = EIte c e1 (go' e2)
-    go' e = e
-
-_splitBranches :: Symbol -> Expr -> [(Expr, Expr)]
```
curiousleo, line 1167: `_splitBranches` was unused.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -1154,42 +1161,36 @@ instance Normalizable Equation where
...
-_splitBranches f = go
-  where
-    go (PAnd es)
-      | f `elem` syms es
-      = go' <$> es
-    go e
-      = [(PTrue, e)]
+    -- Q: Why do we only simplify the expression if f is a free symbol in it?
```
curiousleo: Question 1: Why do we only simplify the expression if f is a free symbol in it?

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -1154,42 +1161,36 @@ instance Normalizable Equation where
...
+    go e | f `elem` syms e = go' e
+    go e                   = e
 
-    go' (PImp c e) = (c, e)
-    go' e          = (PTrue, e)
+    -- | Simplifies a particular logical pattern:
+    -- (c => e1) /\ ((not c) => e2) --> if c then e1 else e2
+    --
+    -- Q: Why do we recurse into the right expression (only)?
```
curiousleo: Question 2: Why do we recurse into the right expression (only)?

src/Language/Fixpoint/Solver/PLE.hs
nikivazou: `normalizeBody f bd` is  only called when `bd` is the definition of `f`. We only need to normalize when `f` is recursive. So, we have this check for efficiency: I believe nothing bad will happen if you normalize anyways, but you do not need to.

src/Language/Fixpoint/Solver/PLE.hs
nikivazou: Again, the expression has a concrete format (that came from Haskell's reflect annotations). The left expression is a conjuction of data constructor checkers, so you do not need to recurse there.

curiousleo:
Thanks for the explanations @nikivazou! I've resolved the questions in the documentation comments.

src/Language/Fixpoint/Solver/PLE.hs
ranjitjhala: Can we add @nikivazou 's remark to answer the question in the comment?

src/Language/Fixpoint/Solver/PLE.hs
ranjitjhala: again, may be good to add @nikivazou 's answer into the comment?

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: I think this has be done in the last commit: eaff89e

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: This has been added in eaff89e a well.

facundominguez:
I merged #520, which I created to fix the conflicts in this branch. Please, let me know if anything looks amiss.

ranjitjhala:
🙏🙏🙏

PR #527, opened 2022-02-18
Test undoANF

Fizzixnerd:
This PR adds tests for `undoANF`.

There are currently 5 tests added: a simple unit test on the empty env, and 4 PBTs using `QuickCheck`.  The properties tests for are:
1. `undoANF` leaves an env without `lq_anf$` vars unchanged, except possibly for simplification
2. `undoANF` removes all references to `lq_anf$` vars from the `Expr`s of the `Reft`s of the `SortedReft`s.

For number 2., we check this in three starting classes of environments: ones with no `lq_anf$`s, ones with only a single level of `lq_anf$` references, and ones with multiple levels of `lq_anf$` references.

Because `simplify` is called in `undoANF`, I have parameterized a bunch of functions on a simplifier.  The reason for this is that `simplify` can reorder the conjuncts, making 1. difficult to test.  Instead, we test `undoANF` without a simplifier (and with a trivial lens) by testing `undoANFSimplifyingWith id id`.

The way this was done (preserving the old functions entirely and introducing new ones) ensures that there is (hopefully!) no change of behavior in the library from the addition of the parameterized functions.  The one exception to this is `undoANFOnlyModified`, because a bug was discovered in it whereby it would return more than just the unmodified bindings due to above mentioned reorderings from `simplify`.

Care was taken to ensure that tests cannot accidentally loop forever by adding a timeout of 1s (tests run in ~0.07s on my machine...), and by carefully creating symbols in a way that should be impossible to form loops.

### DIFF
--- a/src/Language/Fixpoint/Solver/EnvironmentReduction.hs
+++ b/src/Language/Fixpoint/Solver/EnvironmentReduction.hs
@@ -17,6 +17,9 @@ module Language.Fixpoint.Solver.EnvironmentReduction
   , simplifyBooleanRefts
   , undoANF
   , undoANFAndVV
+
+  -- for use in tests
+  , undoANFSimplifyingWith
   ) where
 
 import           Control.Monad (guard, mplus, msum)
@@ -568,26 +571,37 @@ mergeDuplicatedBindings xs =
 -- | Inlines some of the bindings whose symbol satisfies a given predicate.
 --
 -- Only works if the bindings don't form cycles.
-substBindings
-  :: Lens' v SortedReft
+substBindingsSimplifyingWith
+  :: (Expr -> Expr)
+  -> Lens' v SortedReft
   -> (Symbol -> Bool)
   -> HashMap Symbol v
   -> HashMap Symbol v
-substBindings vLens p env =
+substBindingsSimplifyingWith simplifier vLens p env =
     -- Circular program here. This should terminate as long as the
     -- bindings introduced by ANF don't form cycles.
-    let env' = HashMap.map (vLens %~ inlineInSortedReft (srLookup filteredEnv)) env
+    let env' = HashMap.map (vLens %~ inlineInSortedReftSimplifyingWith simplifier (srLookup filteredEnv)) env
         filteredEnv = HashMap.filterWithKey (\sym _v -> p sym) env'
      in env'
   where
     srLookup env' sym = view vLens <$> HashMap.lookup sym env'
 
+substBindings
+  :: Lens' v SortedReft
+  -> (Symbol -> Bool)
+  -> HashMap Symbol v
+  -> HashMap Symbol v
+substBindings = substBindingsSimplifyingWith simplify
+
 -- | Like 'substBindings' but specialized for ANF bindings.
 --
--- Only bindings with prefix lq_anf... might be inlined.
+-- Only bindings with prefix lq_anf$... might be inlined.
 --
+undoANFSimplifyingWith :: (Expr -> Expr) -> Lens' v SortedReft -> HashMap Symbol v -> HashMap Symbol v
+undoANFSimplifyingWith simplifier vLens = substBindingsSimplifyingWith simplifier vLens $ \sym -> anfPrefix `isPrefixOfSym` sym
+
 undoANF :: Lens' v SortedReft -> HashMap Symbol v -> HashMap Symbol v
-undoANF vLens = substBindings vLens $ \sym -> anfPrefix `isPrefixOfSym` sym
+undoANF = undoANFSimplifyingWith simplify
 
 -- | Like 'undoANF' but also inlines VV bindings
 --
@@ -595,31 +609,39 @@ undoANF vLens = substBindings vLens $ \sym -> anfPrefix `isPrefixOfSym` sym
 -- can request to use it in the verification pipeline with
 -- @--inline-anf-bindings@. However, using it in the verification
 -- pipeline causes some tests in liquidhaskell to blow up.
+--
+-- Note: This function simplifies.
 undoANFAndVV :: HashMap Symbol (m, SortedReft) -> HashMap Symbol (m, SortedReft)
 undoANFAndVV = substBindings _2 $ \sym -> anfPrefix `isPrefixOfSym` sym || vvName `isPrefixOfSym` sym
 
--- | Like 'undoANF' but returns only modified bindings
+-- | Like 'undoANF' but returns only modified bindings and **DOES NOT SIMPLIFY**.
 undoANFOnlyModified :: HashMap Symbol (m, SortedReft) -> HashMap Symbol (m, SortedReft)
 undoANFOnlyModified env =
-    let undoANFEnv = undoANF _2 env
+    let undoANFEnv = undoANFSimplifyingWith id _2 env
      in HashMap.differenceWith dropUnchanged env undoANFEnv
   where
     dropUnchanged (_, a) v@(_, b) | a == b = Just v
       | otherwise = Nothing
 
 -- | Inlines bindings in env in the given 'SortedReft'.
-inlineInSortedReft
-  :: (Symbol -> Maybe SortedReft)
+inlineInSortedReftSimplifyingWith
+  :: (Expr -> Expr)
+  -> (Symbol -> Maybe SortedReft)
   -> SortedReft
   -> SortedReft
-inlineInSortedReft srLookup sr =
+inlineInSortedReftSimplifyingWith simplifier srLookup sr =
     let reft = sr_reft sr
-     in sr { sr_reft = mapPredReft (inlineInExpr (filterBind (reftBind reft) srLookup)) reft }
+     in sr { sr_reft = mapPredReft (inlineInExprSimplifyingWith simplifier (filterBind (reftBind reft) srLookup)) reft }
   where
     filterBind b srLookup sym = do
       guard (sym /= b)
       srLookup sym
 
+inlineInSortedReft
+  :: (Symbol -> Maybe SortedReft)
+  -> SortedReft
+  -> SortedReft
+inlineInSortedReft = inlineInSortedReftSimplifyingWith simplify
 
 -- | Inlines bindings given by @srLookup@ in the given expression
 -- if they appear in equalities.
@@ -634,8 +656,8 @@ inlineInSortedReft srLookup sr =
 -- Given a binding like @a : { v | v = e1 }@ and an expression @... a ...@,
 -- this function produces the expression @... e1 ...@ if @v@ does not
 -- appear free in @e1@.
-inlineInExpr :: (Symbol -> Maybe SortedReft) -> Expr -> Expr
-inlineInExpr srLookup = simplify . mapExprOnExpr inlineExpr
+inlineInExprSimplifyingWith :: (Expr -> Expr) -> (Symbol -> Maybe SortedReft) -> Expr -> Expr
+inlineInExprSimplifyingWith simplifier srLookup = simplifier . mapExprOnExpr inlineExpr
   where
     inlineExpr (EVar sym)
       | Just sr <- srLookup sym
@@ -669,6 +691,9 @@ inlineInExpr srLookup = simplify . mapExprOnExpr inlineExpr
     wrapWithCoercion br to e = case exprSort_maybe e of
       Just from -> if from /= to then ECoerc from to e else e
       Nothing -> if br == Ueq then ECst e to else e
+
+inlineInExpr :: (Symbol -> Maybe SortedReft) -> Expr -> Expr
+inlineInExpr = inlineInExprSimplifyingWith simplify
 
 dropECst :: Expr -> Expr
 dropECst = \case
--- a/src/Language/Fixpoint/Solver/EnvironmentReduction.hs
+++ b/src/Language/Fixpoint/Solver/EnvironmentReduction.hs
@@ -580,11 +580,12 @@ substBindingsSimplifyingWith
 substBindingsSimplifyingWith simplifier vLens p env =
     -- Circular program here. This should terminate as long as the
     -- bindings introduced by ANF don't form cycles.
-    let env' = HashMap.map (vLens %~ inlineInSortedReftSimplifyingWith simplifier (srLookup filteredEnv)) env
+    let env' = HashMap.map (vLens %~ sortedReftSimplifier . inlineInSortedReft (srLookup filteredEnv)) env
         filteredEnv = HashMap.filterWithKey (\sym _v -> p sym) env'
      in env'
   where
     srLookup env' sym = view vLens <$> HashMap.lookup sym env'
+    sortedReftSimplifier (RR s r) = RR s (reft (reftBind r) (simplifier (reftPred r)))
 
 substBindings
   :: Lens' v SortedReft
@@ -624,24 +625,17 @@ undoANFOnlyModified env =
       | otherwise = Nothing
 
 -- | Inlines bindings in env in the given 'SortedReft'.
-inlineInSortedReftSimplifyingWith
-  :: (Expr -> Expr)
-  -> (Symbol -> Maybe SortedReft)
-  -> SortedReft
-  -> SortedReft
-inlineInSortedReftSimplifyingWith simplifier srLookup sr =
-    let reft = sr_reft sr
-     in sr { sr_reft = mapPredReft (inlineInExprSimplifyingWith simplifier (filterBind (reftBind reft) srLookup)) reft }
-  where
-    filterBind b srLookup sym = do
-      guard (sym /= b)
-      srLookup sym
-
 inlineInSortedReft
   :: (Symbol -> Maybe SortedReft)
   -> SortedReft
   -> SortedReft
-inlineInSortedReft = inlineInSortedReftSimplifyingWith simplify
+inlineInSortedReft srLookup sr =
+    let reft = sr_reft sr
+     in sr { sr_reft = mapPredReft (inlineInExpr (filterBind (reftBind reft) srLookup)) reft }
+  where
+    filterBind b srLookup sym = do
+      guard (sym /= b)
+      srLookup sym
 
 -- | Inlines bindings given by @srLookup@ in the given expression
 -- if they appear in equalities.
@@ -656,8 +650,8 @@ inlineInSortedReft = inlineInSortedReftSimplifyingWith simplify
 -- Given a binding like @a : { v | v = e1 }@ and an expression @... a ...@,
 -- this function produces the expression @... e1 ...@ if @v@ does not
 -- appear free in @e1@.
-inlineInExprSimplifyingWith :: (Expr -> Expr) -> (Symbol -> Maybe SortedReft) -> Expr -> Expr
-inlineInExprSimplifyingWith simplifier srLookup = simplifier . mapExprOnExpr inlineExpr
+inlineInExpr :: (Symbol -> Maybe SortedReft) -> Expr -> Expr
+inlineInExpr srLookup = simplify . mapExprOnExpr inlineExpr
   where
     inlineExpr (EVar sym)
       | Just sr <- srLookup sym
@@ -691,9 +685,6 @@ inlineInExprSimplifyingWith simplifier srLookup = simplifier . mapExprOnExpr inl
     wrapWithCoercion br to e = case exprSort_maybe e of
       Just from -> if from /= to then ECoerc from to e else e
       Nothing -> if br == Ueq then ECst e to else e
-
-inlineInExpr :: (Symbol -> Maybe SortedReft) -> Expr -> Expr
-inlineInExpr = inlineInExprSimplifyingWith simplify
 
 dropECst :: Expr -> Expr
 dropECst = \case
--- a/src/Language/Fixpoint/Solver/EnvironmentReduction.hs
+++ b/src/Language/Fixpoint/Solver/EnvironmentReduction.hs
@@ -18,6 +18,8 @@ module Language.Fixpoint.Solver.EnvironmentReduction
   , undoANF
   , undoANFAndVV
 
+  , mkSortedReftSimplifier
+
   -- for use in tests
   , undoANFSimplifyingWith
   ) where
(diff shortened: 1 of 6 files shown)


src/Language/Fixpoint/Solver/EnvironmentReduction.hs
```diff
@@ -568,58 +571,77 @@ mergeDuplicatedBindings xs =
...
+  -> Lens' v SortedReft
   -> (Symbol -> Bool)
   -> HashMap Symbol v
   -> HashMap Symbol v
-substBindings vLens p env =
+substBindingsSimplifyingWith simplifier vLens p env =
     -- Circular program here. This should terminate as long as the
     -- bindings introduced by ANF don't form cycles.
-    let env' = HashMap.map (vLens %~ inlineInSortedReft (srLookup filteredEnv)) env
+    let env' = HashMap.map (vLens %~ inlineInSortedReftSimplifyingWith simplifier (srLookup filteredEnv)) env
```
facundominguez: How about a version like this
```suggestion
    let env' = HashMap.map (vLens %~ simplifier . inlineInSorted (srLookup filteredEnv)) env
```
so we don't need to introduce `inlineInSortedReftSimplifyingWith` and `inlineInExprSimplifyingWith`?

The type of `symplifier` would need to change to `SortedReft -> SortedReft`.

tests/tasty/Arbitrary.hs
```diff
@@ -65,9 +77,9 @@ subexprs (PGrad _ _ _ e) = [e]
 subexprs (ECoerc _ _ e)  = [e]
 
 -- TODO: Adjust frequencies
-arbitraryExpr :: Int -> Gen Expr
-arbitraryExpr 0 = oneof [ESym <$> arbitrary, ECon <$> arbitrary, EVar <$> arbitrary]
-arbitraryExpr n = frequency
+arbitraryExpr_ :: Gen Expr -> Int -> Gen Expr
```
facundominguez: Please, consider adding a comment explaining the meaning of the argument generator.

tests/tasty/Arbitrary.hs
```diff
@@ -133,16 +151,17 @@ instance Arbitrary Bop where
 instance Arbitrary SymConst where
   arbitrary = SL <$> arbitrary
 
+-- | Note that we rely below on the property that the Arbitrary instance for
+-- Symbol cannot create lq_anf$ vars.
 instance Arbitrary Symbol where
   arbitrary = (symbol :: Text.Text -> Symbol) <$> arbitrary
 
 instance Arbitrary Text.Text where
```
facundominguez: This instance should probably have a comment saying that it is assumed to only produced text with lowercase letters. You could even make a newtype to make this explicit in the type.

tests/tasty/Arbitrary.hs
```diff
@@ -155,3 +174,109 @@ instance Arbitrary Constant where
...
+newtype NoAnfEnv = NoAnfEnv { unNoAnfEnv :: Env }
+  deriving (Eq, Show, Generic)
+instance Arbitrary NoAnfEnv where
+  arbitrary = NoAnfEnv <$> (arbitraryEnv 4 gen)
+    where
+      -- | Note that this relies on the property that the Arbitrary instance for
+      -- Symbol cannot create lq_anf$ vars.
+      symGen = const arbitrary
+      reftGen = const arbitrary
+      gen = fmap pure . (pairGen symGen reftGen)
```
facundominguez: Why not
```suggestion
      gen _ = (\a b -> [(a, b)]) <$> arbitrary <*> arbitrary
```
?

tests/tasty/Arbitrary.hs
```diff
@@ -155,3 +174,109 @@ instance Arbitrary Constant where
...
+      anfGen' = anfGen symGen (const arbitrary)
+      gen = finalAnfGen (fmap pure . anfGen')
+
+anfGen :: (Int -> Gen AnfSymbol) -> (Int -> Gen SortedReft) -> Int -> Gen (Symbol, SortedReft)
+anfGen symGen sreftGen = \i -> do
+  sym <- unAnfSymbol <$> symGen i
+  sreft <- sreftGen i
+  pure $ (sym, sreft)
+
+finalAnfGen :: (Int -> Gen [(Symbol, SortedReft)]) -> Int -> Gen [(Symbol, SortedReft)]
```
facundominguez: This function would deserve some documentation.

tests/tasty/Arbitrary.hs
```diff
@@ -155,3 +174,109 @@ instance Arbitrary Constant where
...
+
+-- | Env with anf vars that form a list of references.
+newtype ChainedAnfEnv = ChainedAnfEnv { unChainedAnfEnv :: Env }
+  deriving (Eq, Show, Generic)
+instance Arbitrary ChainedAnfEnv where
+  arbitrary = ChainedAnfEnv <$> (arbitraryEnv 8 gen)
+    where
+      gen = finalAnfGen $ chainedAnfGen anfSymNGen
+
+chainedAnfGen :: (Int -> Gen AnfSymbol) -> Int -> Gen [(Symbol, SortedReft)]
```
facundominguez, line 292: This function would deserve some documentation as well.

tests/tasty/Arbitrary.hs
```diff
@@ -155,3 +174,109 @@ instance Arbitrary Constant where
...
+    where
+      symGen = const arbitrary
+      anfGen' = anfGen symGen (const arbitrary)
+      gen = finalAnfGen (fmap pure . anfGen')
+
+anfGen :: (Int -> Gen AnfSymbol) -> (Int -> Gen SortedReft) -> Int -> Gen (Symbol, SortedReft)
+anfGen symGen sreftGen = \i -> do
+  sym <- unAnfSymbol <$> symGen i
+  sreft <- sreftGen i
+  pure $ (sym, sreft)
```
facundominguez: Why not
```suggestion
      anfGen = (\a b -> [(unAnfSymbol a, b)]) <$> arbitrary <*> arbitrary
      gen = finalAnfGen anfGen
```
?

tests/tasty/Arbitrary.hs
```diff
@@ -155,3 +174,109 @@ instance Arbitrary Constant where
...
+
+chainedAnfGen :: (Int -> Gen AnfSymbol) -> Int -> Gen [(Symbol, SortedReft)]
+chainedAnfGen symGen = \i -> do
+  syms <- fmap unAnfSymbol <$> (for [0..i] symGen)
+  let symPairs :: [(Symbol, Symbol)]
+      symPairs = pairs syms
+      anfPairs :: Gen [(Symbol, SortedReft)]
+      anfPairs = traverse (\(sym, prevSym) -> do
+                              otherSym <- arbitrary
+                              (sym,) <$> (RR <$> arbitrary <*> (reft otherSym <$> (PAtom Eq (EVar otherSym) <$> (arbitraryExprInvolving prevSym i))))) symPairs
```
facundominguez: This line is a bit long. I think there's a flipped variant of `traverse` called `for` if that helps tidying the code.

tests/tasty/Arbitrary.hs
```diff
@@ -155,3 +174,109 @@ instance Arbitrary Constant where
...
+-- | Base environment with no declared properties; do not add an Arbitrary
+-- instance to this and instead use newtypes.
+newtype Env = Env { unEnv :: M.HashMap Symbol SortedReft }
+  deriving (Eq, Show, Generic)
+
+-- | Env without anf vars.
+newtype NoAnfEnv = NoAnfEnv { unNoAnfEnv :: Env }
+  deriving (Eq, Show, Generic)
+instance Arbitrary NoAnfEnv where
+  arbitrary = NoAnfEnv <$> (arbitraryEnv 4 gen)
```
facundominguez: Shouldn't the size be taken from the monad, so the caller can decide how large an environment to create? Also, the size here is irrelevant as gen always produces a single pair. Shouldn't we test with larger environments?

tests/tasty/UndoANFTests.hs
```diff
@@ -0,0 +1,76 @@
...
+          toAnfVarSymbols (_, sr) = sortedReftAnfVarSymbols sr
+          anfVarSymbolVisitor :: Visitor.Visitor [Symbol] ()
+          anfVarSymbolVisitor =
+            (Visitor.defaultVisitor :: Visitor.Visitor [Symbol] ()) { Visitor.accExpr = \_ -> \case
+                                       EVar v | isAnfVar v -> [v]
+                                       _ -> []
+                                   }
+          sortedReftAnfVarSymbols = Visitor.fold anfVarSymbolVisitor () mempty
+      in
+        (concatMap toAnfVarSymbols symbolsAndSortedRefts) === []
```
facundominguez: Why not
```suggestion
    checkNoAnfs (Env m) = HashMap.filter (any isAnfVar . syms) m === HashMap.empty
    isAnfVar = isPrefixOfSym anfPrefix
```
?

tests/tasty/UndoANFTests.hs
```diff
@@ -0,0 +1,76 @@
...
+              prop_no_anfs unFlatAnfEnv simpleUndoANFFlatAnfEnv
+          , Q.testProperty "chained anf vars" $
+              prop_no_anfs unChainedAnfEnv simpleUndoANFChainedAnfEnv
+          ]
+      ]
+  where
+    withOptions = localOption (Q.QuickCheckMaxSize 4) . localOption (Q.QuickCheckTests 500)
+
+prop_no_change :: (Q.Arbitrary e, Eq e, Show e) => (e -> e) -> e -> Q.Property
+prop_no_change f e = Q.within 1000000 $ f e === e
```
facundominguez: The timeout could be a bit more generous and it would make little difference to the human. I'd make it 5s or 10s.

facundominguez (commented):
Thanks Matt!

I shared some comments.

src/Language/Fixpoint/Solver/EnvironmentReduction.hs
Fizzixnerd: Sure.  The reason I did it this way was to preserve the code meaning if the function is used elsewhere, which it is in Prettify.  This would mean changing three call sites, changing the behavior of `inlineInSortedReft` (to not simplify), but remove two non-exported function definitions.

I normally would prefer to preserve the meaning of exported functions unless I have a good reason to, and the functions introduced don't escape the file, but I think I'll defer to you in this case as I don't know the priorities of the project well enough to really make a case.

Will change.

tests/tasty/UndoANFTests.hs
Fizzixnerd: Did not know this function exists.  Also, my version was just wrong because it didn't grab only *free* symbols!

tests/tasty/UndoANFTests.hs
Fizzixnerd: Changed.

tests/tasty/UndoANFTests.hs
Fizzixnerd: Changed.

tests/tasty/Arbitrary.hs
Fizzixnerd: Changed.

tests/tasty/Arbitrary.hs
Fizzixnerd: Changed structure to something different.  The anfGens are now inlined and can generate `Env`s of larger sizes.

tests/tasty/Arbitrary.hs
Fizzixnerd: Changed.

tests/tasty/Arbitrary.hs
Fizzixnerd: Changed.

tests/tasty/Arbitrary.hs
Fizzixnerd: Renamed and commented.

src/Language/Fixpoint/Solver/EnvironmentReduction.hs
Fizzixnerd: Changed.

EDIT: Also, I should mention, it was much less painful than I thought it was.  I totally agree with this change now.

tests/tasty/Arbitrary.hs
Fizzixnerd: Done, but I had to remove a couple of shrinkers because they depended on Text having an instance.  They not doing any heavy lifting I think though, because all they did was shorten the length of symbols (as far as I could tell).

tests/tasty/Arbitrary.hs
Fizzixnerd: Changed.

tests/tasty/Arbitrary.hs
```diff
@@ -116,7 +149,7 @@ arbitrarySort 0 = oneof (pure <$> [FInt, FReal, FNum, FFrac])
 arbitrarySort n = frequency
   [ (1, FObj <$> arbitrary)
   , (1, FVar <$> arbitrary)
-  , (1, FFunc <$> arbitrarySort' <*> arbitrarySort')
+  -- , (1, FFunc <$> arbitrarySort' <*> arbitrarySort')
```
Fizzixnerd, line 152: @facundominguez this is the line that was causing the strange errors.  The problem is with a function called `unifySorts` in liquid-fixpoint.  Basically, it calls `die`, but I have no idea how to fix it so I just removed this case from the sorts.  I'm not sure why the sorts are trying to be unified in any case, since we are testing a substitution, not type checking...

tests/tasty/Arbitrary.hs
Fizzixnerd: Changed.

src/Language/Fixpoint/Solver/EnvironmentReduction.hs
```diff
@@ -568,40 +571,54 @@ mergeDuplicatedBindings xs =
...
+substBindingsSimplifyingWith simplifier vLens p env =
     -- Circular program here. This should terminate as long as the
     -- bindings introduced by ANF don't form cycles.
-    let env' = HashMap.map (vLens %~ inlineInSortedReft (srLookup filteredEnv)) env
+    let env' = HashMap.map (vLens %~ sortedReftSimplifier . inlineInSortedReft (srLookup filteredEnv)) env
         filteredEnv = HashMap.filterWithKey (\sym _v -> p sym) env'
      in env'
   where
     srLookup env' sym = view vLens <$> HashMap.lookup sym env'
+    sortedReftSimplifier (RR s r) = RR s (reft (reftBind r) (simplifier (reftPred r)))
```
facundominguez: I think that you intended to remove the `simplify` call from `inlineInExpr` which would be redundant now, and update the call sites of `inlineInExpr` and `inlineInSortedReft`.

Fizzixnerd: Yes, of course.  I think I confused myself earlier!  Changed.

src/Language/Fixpoint/Solver/EnvironmentReduction.hs
```diff
@@ -585,7 +587,10 @@ substBindingsSimplifyingWith simplifier vLens p env =
      in env'
   where
     srLookup env' sym = view vLens <$> HashMap.lookup sym env'
-    sortedReftSimplifier (RR s r) = RR s (reft (reftBind r) (simplifier (reftPred r)))
+    sortedReftSimplifier = mkSortedReftSimplifier simplifier
+
+mkSortedReftSimplifier :: (Expr -> Expr) -> SortedReft -> SortedReft
+mkSortedReftSimplifier simplifier (RR s r) = RR s (reft (reftBind r) (simplifier (reftPred r)))
```
facundominguez, line 593: Lets remember to remove this function in another PR.
`SortedReft` has an instance of the class `Fixpoint`, and the class `Fixpoint` has a `simplify` method. However the instance for `SortedReft` defines `simplify` as `id` instead of something like `mkSortedReftSimplifier` which, arguably, would be a more reasonable implementation.

facundominguez (approved):
LGTM!

tests/tasty/Arbitrary.hs
facundominguez: These environment only have one binding still, right? Can we make `gen` pay attention to the size parameter?

tests/tasty/Arbitrary.hs
```diff
@@ -76,26 +93,42 @@ arbitraryExpr n = frequency
...
+arbitraryEqualityConstraint = oneof [pure Eq, pure Ueq]
+
+-- | Generates an expression that involves the given Symbol in an Eq or Ueq
+-- PAtom with an arbitrary non-lq_anf$ EVar.
+arbitraryExprInvolving :: Symbol -> Int -> Gen Expr
+arbitraryExprInvolving sym = arbitraryFiniteExpr . pure $ EVar sym
 
 instance Arbitrary KVar where
   arbitrary = KV <$> arbitrary
-  shrink (KV s) = KV <$> genericShrink s
```
facundominguez, line 98: Why is this removed? I just tried adding it back and found no obvious failures.

Fizzixnerd: It will not compile for me.

```iquid-fixpoint  > [ 6 of 10] Compiling Arbitrary
liquid-fixpoint  > 
liquid-fixpoint  > /home/fizzixnerd/src/tweag/internal/liquidhaskell/liquid-fixpoint/tests/tasty/Arbitrary.hs:132:26: error:
liquid-fixpoint  >     • No instance for (Arbitrary Text.Text)
liquid-fixpoint  >         arising from a use of ‘genericShrink’
liquid-fixpoint  >     • In the second argument of ‘(<$>)’, namely ‘genericShrink s’
liquid-fixpoint  >       In the expression: KV <$> genericShrink s
liquid-fixpoint  >       In an equation for ‘shrink’: shrink (KV s) = KV <$> genericShrink s
liquid-fixpoint  >     |
liquid-fixpoint  > 132 |   shrink (KV s) = KV <$> genericShrink s
liquid-fixpoint  >     |                          ^^^^^^^^^^^^^^^
liquid-fixpoint  > 
```

src/Language/Fixpoint/Solver/EnvironmentReduction.hs
Fizzixnerd: Alright, sounds good.

tests/tasty/Arbitrary.hs
Fizzixnerd: Sorry, I changed it for FlatAnfEnvs, I think.  I have now changed it for `NoAnfEnv` too.

tests/tasty/Arbitrary.hs
facundominguez: I think I didn't add back exactly the same code.

tests/tasty/UndoANFTests.hs
```diff
@@ -0,0 +1,72 @@
...
+          [ Q.testProperty "no anf vars" $
+              prop_no_anfs unNoAnfEnv simpleUndoANFNoAnfEnv
+          , Q.testProperty "single-level anf vars" $
+              prop_no_anfs unFlatAnfEnv simpleUndoANFFlatAnfEnv
+          , Q.testProperty "chained anf vars" $
+              prop_no_anfs unChainedAnfEnv simpleUndoANFChainedAnfEnv
+          ]
+      ]
+  where
+    withOptions = localOption (Q.QuickCheckMaxSize 8) . localOption (Q.QuickCheckTests 500)
```
facundominguez: Please, use `adjustOption` here so we have a chance to modify these parameters in the command line.

Maybe
```suggestion
    withOptions = adjustOption (min (Q.QuickCheckMaxSize 8)) . adjustOption (max (Q.QuickCheckTests 500))
```

facundominguez: I updated the suggestion above.

tests/tasty/Arbitrary.hs
facundominguez: Sometimes `inlineInExpr` would need to insert sort coercions when inlining a binding which doesn't have exactly the same sort as expected at the place in which the binding is referenced. Here `inlineInExpr` needs to determine what the actual and the expected sorts are, and there is where sort inference is involved, which sometimes seems to require unification.

Are you sure no other code relies on the case that you are leaving out? If you are, then this is fine. But otherwise, it would be preferable to just filter this case at the places where you use it.

Also, maybe arbitrary sorts are not necessary and a single sort could be used, say `FInt`.

tests/tasty/Arbitrary.hs
```diff
@@ -131,27 +164,148 @@ instance Arbitrary Bop where
...
+--
+-- This list is in an acceptable form to be passed to `finalAnfGen` to close the
+-- loop.
+chainedAnfGen :: (Int -> Gen AnfSymbol) -> Int -> Gen [(Symbol, SortedReft)]
+chainedAnfGen _ 0 = pure []
+chainedAnfGen symGen n = do
+  syms <- fmap unAnfSymbol <$> (for [1..n+1] symGen)
+  finalSym <- arbitrary
+  let symPairs :: [(Symbol, Symbol)]
+      symPairs = pairs syms ++ [(last syms, finalSym)]
```
facundominguez:
```suggestion
      symPairs = pairs (syms ++ [finalSym])
```

facundominguez:
I found some more things to fix after approving. I think they are all easy but let me know if you'd like deferring any of those.

tests/tasty/UndoANFTests.hs
Fizzixnerd: Making this change causes the NoAnfEnv tests to fail due to timeout after 10 tests.  Changing the first `adjustOption` to `const (Q.QuickCheckMaxSize 8)` (which I believe is equivalent to the old `localOption`) makes it run in time again.  I think the default `QuickCheckMaxSize` is larger than 8.  I could set this value to lower on the command line, or do you think it's okay to use `localOption` again?

tests/tasty/Arbitrary.hs
Fizzixnerd: Good catch!  Changed.

tests/tasty/UndoANFTests.hs
facundominguez: I edited the suggestion to use the smaller of the current size and `8`. Would that have tests run snappy?

Fizzixnerd: That fixes it.  Changed.

tests/tasty/Arbitrary.hs
Fizzixnerd: I like the idea of using a single sort, actually, because that would make the test cases much smaller to read.

PR #530, opened 2022-02-23
Make Envs generated only IntSortedRefts

### LINKED ISSUE #529: Use `IntSortedReft`s instead of arbitrary `SortedReft`s in test cases for `undoANF`
Sometimes `inlineInExpr` would need to insert sort coercions when inlining a binding which doesn't have exactly the same sort as expected at the place in which the binding is referenced. Here `inlineInExpr` needs to determine what the actual and the expected sorts are, and there is where sort inference is involved, which sometimes seems to require unification.

Are you sure no other code relies on the case that you are leaving out? If you are, then this is fine. But otherwise, it would be preferable to just filter this case at the places where you use it.

Also, maybe arbitrary sorts are not necessary and a single sort could be used, say `FInt`.

_Originally posted by @facundominguez in https://github.com/ucsd-progsys/liquid-fixpoint/pull/527#discussion_r813010129_

Fizzixnerd:
Closes #529.

### DIFF
--- a/tests/tasty/Arbitrary.hs
+++ b/tests/tasty/Arbitrary.hs
@@ -149,7 +149,7 @@ arbitrarySort 0 = oneof (pure <$> [FInt, FReal, FNum, FFrac])
 arbitrarySort n = frequency
   [ (1, FObj <$> arbitrary)
   , (1, FVar <$> arbitrary)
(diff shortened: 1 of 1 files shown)


facundominguez (approved):
LGTM. Thanks!

PR #531, opened 2022-02-23
SortedReft Fixpoint instance fix

### LINKED ISSUE #528: Make `SortedReft`'s `Fixpoint` instance equal to something like `mkSortedReftSimplifier`
Lets remember to remove this function in another PR.
`SortedReft` has an instance of the class `Fixpoint`, and the class `Fixpoint` has a `simplify` method. However the instance for `SortedReft` defines `simplify` as `id` instead of something like `mkSortedReftSimplifier` which, arguably, would be a more reasonable implementation.

_Originally posted by @facundominguez in https://github.com/ucsd-progsys/liquid-fixpoint/pull/527#discussion_r812928114_

Fizzixnerd:
Closes #528.

### DIFF
--- a/src/Language/Fixpoint/Solver/EnvironmentReduction.hs
+++ b/src/Language/Fixpoint/Solver/EnvironmentReduction.hs
@@ -18,8 +18,6 @@ module Language.Fixpoint.Solver.EnvironmentReduction
   , undoANF
   , undoANFAndVV
 
-  , mkSortedReftSimplifier
-
   -- for use in tests
   , undoANFSimplifyingWith
   ) where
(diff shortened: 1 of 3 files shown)


facundominguez:
Looks good. How about `substBindingsSimplifyingWith` takes a function to simplify `SortedReft`s instead of `Expr`s?

Fizzixnerd:
> Looks good. How about `substBindingsSimplifyingWith` takes a function to simplify `SortedReft`s instead of `Expr`s?

Changed.

facundominguez (approved):
Thanks Matt!

PR #532, opened 2022-02-28
Cleanups and optimizations for PLE

facundominguez:
Cuts the time by half for a couple of benchmarks. It reduces the time it takes to verify stitch-lh by a couple of seconds.

I've tested that tests aren't broken in lh when run with `--no-interpreter`.

### DIFF
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -478,9 +478,8 @@ getAutoRws γ ctx =
 
 evalOne :: Knowledge -> EvalEnv -> ICtx -> Int -> Expr -> IO (EvAccum, FuelCount)
 evalOne γ env ctx i e | i > 0 || null (getAutoRws γ ctx) = do
-    ((e', _), st) <- runStateT (eval γ ctx NoRW e) (env { evFuel = icFuel ctx })
-    let evAcc' = if mytracepp ("evalOne: " ++ showpp e) e' == e then evAccum st else S.insert (e, e') (evAccum st)
-    return (evAcc', evFuel st)
+    st <- execStateT (eval γ ctx NoRW e) (env { evFuel = icFuel ctx })
+    return (evAccum st, evFuel st)
 evalOne γ env ctx _ e = do
   env' <- execStateT (evalREST γ ctx rp) (env { evFuel = icFuel ctx })
   return (evAccum env', evFuel env')
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -447,7 +447,8 @@ type EvAccum = S.HashSet (Expr, Expr)
 --------------------------------------------------------------------------------
 data EvalEnv = EvalEnv
   { evEnv      :: !SymEnv
-  , evAccum    :: EvAccum
+  , evAccum    :: EvAccum -- ^ A cache of equalities between expressions that are
+                          -- known to hold.
   , evFuel     :: FuelCount
 
   -- REST parameters
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -243,13 +243,12 @@ evalCandsLoop cfg ictx0 ctx γ env = go ictx0 0
                   let env' = env { evAccum = icEquals ictx <> evAccum env
                                  , evFuel  = icFuel   ictx
                                  }
-                  (ictx', evalResults)  <- do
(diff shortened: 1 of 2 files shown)


facundominguez:
Here are some performance comparisons
perf

I made the commits small and self-contained so they are easy to review one at a time.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -932,12 +964,16 @@ data Knowledge = KN
...
-    then return False
-    else knPreds γ (knContext γ) (knLams γ) e
+isValid :: IORef (M.HashMap Expr Bool) -> Knowledge -> Expr -> IO Bool
+isValid cacheRef γ e = do
+    smtCache <- readIORef cacheRef
+    case M.lookup e smtCache of
+      Nothing -> do
+        b <- knPreds γ (knContext γ) (knLams γ) e
+        when b $
+          writeIORef cacheRef (M.insert e True smtCache)
```
ranjitjhala, line 974: I am puzzled by one thing: the `e := True` is _written_  to the cache depending on the result at a _particular_ environment `γ`. However,
it is then later _read_ from the cache independent of the environment.

Why is this sound? 

For example, in a particular `γ` we may have `x == y` which will tell us that the two terms are equal (under `γ`) but the terms may not be equal in _other_ contexts? 

Perhaps I'm missing something?

facundominguez: Perhaps there is unsoundness. In my analysis of the code, it looks like `γ` is never trimmed while the cached smt queries are kept.

For instance, if `γ` is trimmed when switching sibling nodes of the Trie, it wouldn't be a problem because the cache isn't shared across siblings.

Said otherwise, let's look for unsoundness here. When are equalities retired from the environment `γ`?

ranjitjhala: Ah, so equalities are never retired as such, but `γ` only "grows" according to the call stack. [I wonder if I can draw a picture.]

A constraint that looks like (with equalities in `p1`, `p2`, `p3` ...) 

```
forall x1:p1 => 
      forall x2:p2 => q
  /\  forall x3:p3 => r
```

Generates a trie like 

```
x1:p1
`------ x2:p2 
|             `----------- q
|            
`------ x3:p3
             `------------ r
```            

And the (old) PLE traverses the trie tracking the environment
explicitly as a parameter with `[]`, then `[x1:p1]` then `[x1:p1, x2:p2]` *then back to* `[x1:p1]` then `[x1:p1, x3:p3]` ...

It seemed to me that your `cache` is now "global" so the equalities from `p2` would also be in the context when we're at `r` ... 

However, you write that

> because the cache isn't shared across siblings.

Can you clarify why that happens?

ranjitjhala: [In particular, previously the `ICtx` holds the learned information at "each node" of the trie, and we explicitly pass it around so that we have "different" `ICtx` at different siblings]

facundominguez: >> because the cache isn't shared across siblings.
>
> Can you clarify why that happens?

At the point in the history that this change is introduced (5c8beb3), every call to `evalCandsLoop` starts with an empty cache (which comes in the `env0` parameter). This means the cache is only used to discover the equalities of a single node in the trie. It also makes some sense to do the same with `evAccum` since equalities discovered in parents are still valid in descendents.

This is changed later on in 75c49b16, where the cache is passed to all of the descendents of a node. However, on each path of the trie, different copies of the cache are used, so queries added to one copy don't affect the others.

facundominguez: > [In particular, previously the ICtx holds the learned information at "each node" of the trie, and we explicitly pass it around so that we have "different" ICtx at different siblings]

I'm fuzzy still at whether to put some fields in `ICtx` or `EvalEnv`. In 8f942ba5 I'm removing `icFuel` for instance, which the code struggled to keep in sync between both `ICtx` and `EvalEnv`.

Perhaps we could reserve `ICtx` for cross-node information and `EvalEnv` for node-local information, but then it puzzles me that an `EvalEnv` is provided from outside `evalCandsLoop` and `ple1` as a parameter :).

ranjitjhala: Yes you’re right - but let’s merge these in and then we can look into that separately?

ranjitjhala (approved):

ranjitjhala:
@nikivazou did you want to take a look or shall we merge?

nikivazou:
Looks great!!!! Merging!

PR #534, opened 2022-03-01
Fix comment about what filterValid does

shingarov:
The comment above filterValid, talks about the pairs (x1,q1) which
constitute Cand.  This order is flipped from what it actually is,
causing a slight confusion for readers trying to understand the code.

### DIFF
--- a/src/Language/Fixpoint/Solver/Monad.hs
+++ b/src/Language/Fixpoint/Solver/Monad.hs
@@ -183,7 +183,7 @@ filterRequired = error "TBD:filterRequired"
 -}
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


ranjitjhala (approved):

ranjitjhala:
Thank you very much for this correction!!!

shingarov:
I wonder what caused the CI to fail, and whether I should do anything to progress this further.

ranjitjhala:
Wow that’s odd will look ASAP!

facundominguez:
I think it is safe to merge this fix. To fix CI we probably need to change flags to apt. https://github.com/ucsd-progsys/liquid-fixpoint/pull/535

PR #537, opened 2022-03-24
Implement the flag --no-interpreter

facundominguez:
Otherwise, one has to use `--useInterpreter=False` which is a bit clunky and isn't documented in the help message.


facundominguez:
Hello! Given that this looks rather uncontroversial and would be helpful, I would be merging it later this week. But let me know if you'd like me to wait.

ranjitjhala:
Please do merge, thanks!

Issue #536, opened 2022-03-11
'ple_sum_fuel.5.smt2' tests fail after 'rest-rewrite' bump


waivio:
I ran into some problems, I worked with @facundominguez, to bump the
`rest-rewrite` library to use a more modern version of `hashable`.
During this effort @facundominguez recomended that I attempt to fix
the upstream dependencies of the library.  The only library dependent on
`rest-rewrite` is `liquid-fixpoint`, and that is convenient because I
would also like to do some ground-work to get it working with
`ghc-9.0.2`.  When I made the simple change to bump the `hashable`
version to `hashable-1.3.5.0` and update the `rest-rewrite` commit to
the final version, the `liquid-fixpoint` tests failed in two spots:

```
      ple_sum_fuel.5.smt2:    FAIL (0.04s)
        tests/test.hs:118:
        Wrong exit code
        expected: ExitFailure 1
         but got: ExitSuccess
        Use -p '/horn-neg-el.ple_sum_fuel.5.smt2/' to rerun this test only.
```
and
```
      ple_sum_fuel.5.smt2:    FAIL (0.04s)
        tests/test.hs:118:
        Wrong exit code
        expected: ExitFailure 1
         but got: ExitSuccess
        Use -p '/horn-neg-na.ple_sum_fuel.5.smt2/' to rerun this test only.

```

Apparently the test is expecting failure:
```
    , testGroup "horn-pos-el" <$> dirTests elimCmd   "tests/horn/pos"  []          ExitSuccess
    , testGroup "horn-neg-el" <$> dirTests elimCmd   "tests/horn/neg"  []          (ExitFailure 1)
    , testGroup "horn-pos-na" <$> dirTests nativeCmd "tests/horn/pos"  []          ExitSuccess
    , testGroup "horn-neg-na" <$> dirTests nativeCmd "tests/horn/neg"  []          (ExitFailure 1)
```
And now it is passing instead of `(ExitFailure 1)`.

Apparently the two tests are the same, but with different values of the
`--fuel` value.

```
(constant sum  (func(0, [int, int])))

(define sum(n : int) : int = { if (n <= 0) then (0) else (n + sum (n-1)) })

(constraint 
   (forall ((x int) ((5 <= x) && (0 <= (sum (x-5))))) 
       ((15 <= (sum x)))
   )
)
```

I am having trouble understanding this case, because if I plug in either
`5` or `6` into the constraint, both would yield true.

[...]

PR #545, opened 2022-04-08
Bump the version of rest-rewrite and fix failing tests

waivio:
Bump the version of rest-rewrite and fix failing tests by reducing the fuel by two.  This PR is related to Issue #536 

Please review and let me know what changes I need to make.

Thanks.
Nathan.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,6 +1,9 @@
 
-resolver: lts-18.14
-compiler: ghc-8.10.7
+resolver: nightly-2022-04-02
+# resolver: nightly-2022-03-10
+# resolver: lts-18.14
+# compiler: ghc-9.0.2
+# compiler: ghc-8.10.7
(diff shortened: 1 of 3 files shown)


stack.yaml
```diff
@@ -1,6 +1,9 @@
 
-resolver: lts-18.14
-compiler: ghc-8.10.7
+resolver: nightly-2022-04-02
```
facundominguez: Thanks @waivio.

Is changing the compiler to 9.2.2 necessary? That is the compiler of `nightly-2022-04-02`. And perhaps more importantly, is changing the snapshot necessary?

LH is using  `lts-18.27` and it has built `hashable-1.3.5.0` fine in my tests.

waivio:
Sorry it was not required.
Just testing out forward compatibility.

Will correct, when possible.

Issue #552, opened 2022-04-15
Which hlint hints do we care about?


philderbeast:
Running hlint-3.3.6, I'm finding 481 hints when run on `.` and 462 hints when run on `src/`:

# Warnings currently triggered by your code
- [ ] Avoid lambda, 4 hints
- [ ] Eta reduce, 30 hints
- [x] Evaluate, 1 hint
- [x] Fix pragma markup, 38 hints
- [x] Functor law, 3 hints
- [x] Move brackets to avoid $, 8 hints
- [x] Redundant $, 20 hints
- [x] Redundant <$>, 29 hints
- [x] Redundant bang pattern, 33 hints
- [x] Redundant bracket, 198 hints
- [x] Redundant flip, 1 hint
- [x] Redundant guard, 2 hints
- [x] Redundant if, 1 hint
- [x] Redundant lambda, 3 hints
- [x] Redundant return, 1 hint
- [x] Unused LANGUAGE pragma, 29 hints
- [ ] Use $>, 7 hints
- [x] Use ++, 1 hint
- [x] Use /=, 1 hint
- [ ] Use <$, 2 hints
- [x] Use <$>, 10 hints
- [x] Use all, 2 hints
- [x] Use camelCase, 2 hints
- [x] Use concatMap, 3 hints
- [x] Use const, 1 hint
- [x] Use elem, 1 hint
- [x] Use first, 1 hint
- [x] Use fold, 1 hint
- [x] Use gets, 9 hints
- [x] Use guards, 3 hints
- [x] Use if, 1 hint
- [ ] Use infix, 3 hints
- [x] Use map, 2 hints
- [x] Use map once, 1 hint
- [x] Use mapM_, 1 hint
- [x] Use mapMaybe, 4 hints
- [x] Use maybe, 7 hints
- [x] Use newtype instead of data, 3 hints
- [x] Use notElem, 2 hints
- [x] Use record patterns, 5 hints
- [x] Use second, 1 hint
- [x] Use section, 3 hints
- [x] Use tuple-section, 1 hint
- [x] Use void, 2 hints

Which of these do we care about?

nikivazou:
I am not using hlint (seems I should) but all these edits look good! 
If you want to apply them and pull request, please do!

philderbeast:
That's all hlint hints followed except for the following that we're ignoring in `.hlint.yaml`:

```yaml
# The following are kept for usability and style sake
[...]

PR #553, opened 2022-04-15
Action to run hlint.

philderbeast:
Adds a github action to run hlint on the `src/` folder with all hints ignored, those listed in #552.


nikivazou:
Thanks!!!

PR #554, opened 2022-04-18
Remove unused LANGUAGE pragmas.

philderbeast:
Tentative first step following hlint suggestions from #552. Also adds badge for hlint github action and runs hlint on other folders with *.hs source files.


ranjitjhala:
Thank you very much!!!

PR #556, opened 2022-04-21
Use cabal generation to check cabal file.

philderbeast:
I used [hpack-dhall](https://hackage.haskell.org/package/hpack-dhall) to generate `liquid-fixpoint.cabal` and then compared the existing file with the generated one. To enable me to more easily do the comparision I reordered some of the fields but tried to keep this to a minimum (sorry if the reformatting is annoying).

The changes are:

* homepage goes straight to the #readme
* added a bug-reports field
* found modules that were not `exposed-modules` or `other-modules` in the library
    ```
    Language.Fixpoint.Congruence.Closure
    Language.Fixpoint.Congruence.Types
    Language.Fixpoint.Types.Binary
    ```
* modules and dependencies are alphabetically sorted

I didn't change the copyright year but see that it is `2010-17`.


philderbeast:
Do we need the `Binary.hs` module still? It seems very empty.

```
-- | We need Binary instances as they are used to serialize specs in LH...
--
module Language.Fixpoint.Types.Binary where

import qualified Data.Binary as B

import Language.Fixpoint.Types as F
```

nikivazou:
Not entirely sure if the files you are deleting are dead code, so better if @ranjitjhala confirms.

philderbeast:
@ranjitjhala are you OK if we delete those 3 modules?

ranjitjhala:
Does LH build without those?

facundominguez:
I don't see them imported anywhere, so they look safe to delete.

ranjitjhala:
👍🏽

facundominguez:
Merging this in #601.

Issue #564, opened 2022-04-28
One of tested-with?


philderbeast:
Are we testing with all `tested-with` versions of GHC?

https://github.com/ucsd-progsys/liquid-fixpoint/blob/db2c3c7ce116cf953d361e04e918cd3be0732c31/liquid-fixpoint.cabal#L24

I've looked at `.circleci/config.yml` and it seems to be doing one `stack build` so picking up `ghc-8.10.7` from `resolver: lts-18.27`.

I'm testing with Z3 and multiple GHC versions as a github action for [plugins-for-blobs](https://github.com/BlockScope/plugins-for-blobs) if that helps.

philderbeast:
Could we go with the latest in each series (dropping < 8):

```diff
- tested-with: GHC == 7.10.3, GHC == 8.0.1, GHC == 8.4.3, GHC == 8.6.4, GHC == 8.10.1, GHC == 8.10.7
+ tested-with: GHC == 8.0.2, GHC == 8.4.4, GHC == 8.6.5, GHC == 8.8.4, GHC == 8.10.7, GHC == 9.0.2, GHC == 9.2.2
```

ranjitjhala:
I think it’s fine to drop everything < 8.8

facundominguez:
Closing this since we merged #591. Feel free to reopen if there is more to do.

Issue #538, opened 2022-04-07
Build with -Werror -Wall


facundominguez:
I've tripped a couple of times over unintended name captures. It would be for the better to enable `-Werror` `-Wall` when building all modules.

`stack` has the `--pedantic` flag to turn on these.

These flags are usually specified for development builds or in CI, but not when building `liquid-fixpoint` as a dependency (this would interfere with the user ability to install it).

`liquid-fixpoint.cabal` already has `-Werror` behind a `devel` flag. For builds with `cabal-install` this could be augmented to use `-Wall` as well.

ranjitjhala:
Excellent idea!!

philderbeast:
Dropping `GHC == 7.10.3" would allow us to use the newer warning syntax.

https://github.com/ucsd-progsys/liquid-fixpoint/blob/db2c3c7ce116cf953d361e04e918cd3be0732c31/liquid-fixpoint.cabal#L148

> In GHC < 8 the syntax for -W<wflag> was -fwarn-<wflag> (e.g. -fwarn-incomplete-patterns). This spelling is deprecated, but still accepted for backwards compatibility. Likewise, -Wno-<wflag> used to be fno-warn-<wflag> (e.g. -fno-warn-incomplete-patterns).
> SOURCE: [using warnings](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/using-warnings.html)

ranjitjhala:
Yes please thank you!

philderbeast:
I'm most of the way there with this one, can it be assigned to me please?

philderbeast:
@facundominguez is this complete with #591?

Are we going to try for a stretch goal of having fewer `{-# OPTIONS_GHC -Wno-name-shadowing #-}`?

facundominguez:
> Are we going to try for a stretch goal of having fewer {-# OPTIONS_GHC -Wno-name-shadowing #-}?

This would be helpful to have eventually. I'd leave this issue open until then.

Issue #557, opened 2022-04-21
SortChecl: lookup in Empty Env


philderbeast:
Found on #555.

```
> cabal test all
...
  undoANFSimplifyingWith id id
    id on empty env:                           OK
    id when env contains no lq_anf$* bindings: OK (0.10s)
      +++ OK, passed 500 tests.
    zero anf vars left afterwards, starting with:
      no anf vars:                             OK (0.05s)
        +++ OK, passed 500 tests.
      single-level anf vars:                   FAIL
...
        Use --quickcheck-replay=407282 --quickcheck-max-size=8 to reproduce.
        Use -p '/single-level anf vars/' to rerun this test only.
```

How I went about checking if it is reproducible (after adding hidden dependencies to the tasty test-suite so that I could run in it the REPL):

```
> stack repl liquid-fixpoint:tasty
...
:main --quickcheck-replay=407282 --quickcheck-max-size=8 "--pattern=single-level anf vars"
Tests
  undoANFSimplifyingWith id id
    zero anf vars left afterwards, starting with:
      single-level anf vars: FAIL (0.01s)
        *** Failed! Exception: 'Error [Error1 {errLoc = SS {sp_start = SourcePos {sourceName = "", sourceLine = Pos 1, sourceColumn = Pos 1}, sp_stop = SourcePos {sourceName = "", sourceLine = Pos 1, sourceColumn = Pos 1}}, errMsg = SortChecl: lookup in Empty Env }]' (after 32 tests):
        FlatAnfEnv {unFlatAnfEnv = Env {unEnv = fromList [("lq_anf$necmvtlyy",{mqgfcf : int | [(mqgfcf ~~
                            (&& [(vvhvaqbcfbr => ybzyzkjxclz);
                                 ("gltgdkgwdes" => "aouxfb");
                                 tabs vylzqqncup xpeqdfk] mod
                               || [exists [lrafyapmrw : @(-4);
                                           utznfu : Sdqrkejej;
                                           frgqwwyhggzp : func(0 , [func(1 , [@(3)]); Gdauhxtm])]
                                          . "xzecjwsu";
                                   forall [feaqzm : @(5);
                                           dszfdv : func(0 , [@(-5); rprjq; Porzvdevj])]
                                          . otygzfmkmv;
                                   tabs 3.0845977044261113 wrbzyynr]))]}),("lq_anf$gzgbktjtf",{vtnqjgfz : int | [(vtnqjgfz ~~
[...]

PR #565, opened 2022-04-28
Test matrix with updated tested-with range.

philderbeast:
Add github actions for testing with cabal and stack, for #564 and before getting started on #538.

I didn't remove the circleci test. I did however add `stack.yaml`. That won't bother the github stack action because it picks up its own `stack/stack-8.10.7.yaml` but it might bother the circleci test.


philderbeast:
The top of the readme rendered:

Screen Shot 2022-04-28 at 11 59 35 AM

philderbeast:
The "single-level anf vars" test #557 is failing often enough now to upset merge checks.

facundominguez:
I think we can drop testing with ghc-8.8.4 and ghc-9.0.2. @nikivazou, @ranjitjhala, any thoughts?

facundominguez:
I'm merging this in #591. I'm removing 8.8.4 and 9.0.2 jobs, but we can add those later if they are needed.

ranjitjhala:
884 Sounds good but why 9?

facundominguez:
There's a job for 9.2.2 to check that LF stays on top of GHC changes. Do you still want to check the intermediate 9.0.2?

ranjitjhala:
I think 9.2.2 is quite sufficient! Thank you!

Issue #539, opened 2022-04-07
Collect performance statistics for liquid-fixpoint


facundominguez:
There are some statistics collected currently, and spread on different flags (`--elimstats --solverstats --stats`).

There are a few things we could do to make them more useful:
1. Have one flag to dump them all.
2. Improve the output to give more context on what they mean.
3. Add a few more stats:
   * how much time is spent in elim, the interpreter ple, and the regular PLE.
   * how much time is spent doing calls to the SMT solver on each algorithm (and how many calls).

Fizzixnerd:
I'd like to take a look at this issue.

Fizzixnerd:
I've added a `--profile` flag so far, and used it to control the dumping of profiling stats.  After a conversation with @facundominguez , I've decided that the way to go is to just do the most "obvious" thing and wrap calls to the ple and solver in a simple `get time, execute action, get time, subtract two times and print, return action result` kind of workflow.  This may not catch everything, since Haskell is of course lazy in the inner pure computations still, possibly.  But it will at least let us know how much time is being spent talking to the SMT, so I have called this helper function `profileIO`.

Fizzixnerd:
regarding having one flag to dump them all: note that `--stats` actually doesn't run the solver, it _only_ prints the stats.  Therefore, a somewhat larger refactor may be required to achieve this goal; it might end up being easy, however.

facundominguez:
> 3. Add a few more stats:
>      * how much time is spent in elim, the interpreter ple, and the regular PLE.
>      * how much time is spent doing calls to the SMT solver on each algorithm (and how many calls).

I'm considering [timestats](https://github.com/tweag/timestats) to solve some of these.

PR #568, opened 2022-05-09
Add profile flag and profile ple/interpreter/solver

Fizzixnerd:
Related to #539.

Adds a `--profile` flag, which when provided will output some basic profiling information, such as how much time was spent in the IO of the PLE, PLE interpreter, and overall in the solver.

### DIFF
--- a/liquid-fixpoint.cabal
+++ b/liquid-fixpoint.cabal
@@ -109,6 +109,7 @@ library
                     Language.Fixpoint.Utils.Progress
                     Language.Fixpoint.Utils.Statistics
                     Language.Fixpoint.Utils.Trie
+                    Language.Fixpoint.Utils.Profile
                     Text.PrettyPrint.HughesPJ.Compat
   other-modules:    Paths_liquid_fixpoint
   hs-source-dirs:   src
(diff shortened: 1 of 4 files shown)


src/Language/Fixpoint/Utils/Profile.hs
```diff
@@ -0,0 +1,13 @@
...
+import Data.Time.Clock
+import Control.Monad
+import Control.Monad.IO.Class
+
+profileIO :: MonadIO m => Bool -> String -> m a -> m a
+profileIO shouldProfile header m = do
+  beforeTime <- liftIO getCurrentTime
+  ret <- m
+  afterTime <- liftIO $ getCurrentTime
+  when shouldProfile $ liftIO $ putStrLn $ header <> show (afterTime `diffUTCTime` beforeTime)
```
facundominguez, line 12: I think it will work better to collect these measures somewhere and print them at the end of the verification. Also, I imagine that we will want to add up a few of these measures when we call multiple times to the same function.

liquid-fixpoint.cabal
```diff
@@ -144,6 +145,7 @@ library
                   , unordered-containers
                   , aeson
                   , rest-rewrite >= 0.1.1
+                  , time
```
facundominguez, line 148: Probably we should be using the `clock` package instead of `time` since we don't care of time zones or days.

facundominguez:
Since this PR, [timestats](https://github.com/tweag/timestats) was released, which could be simpler to use.

PR #582, opened 2022-05-28
nix flake for liquid-fixpoint

plredmond:
This is the nix flake I maintain [over here](https://github.com/ucsd-progsys/liquid-fixpoint/compare/develop...plredmond:nix-flake#files_bucket).

Nix flakes allow nix users to easily build and depend on other projects, without integrating the build instructions into nixpkgs. It's like a project-format for nix.

This probably requires review and improvement by other LH users who are nix users.

* [ ] Update documentation in the [README](https://github.com/ucsd-progsys/liquid-fixpoint/blob/deacde796cc666d7fbeec0fe23f1d587915401be/README.md#how-to-build-and-install) about building with stack and/or nix.
* [ ] Check for references to deleted nix files and provide replacements using [flake-compat](https://github.com/edolstra/flake-compat)


plredmond:
I should probably squash this history out.

facundominguez:
Thanks @plredmond. I see this PR is removing the setup to build with nix and stack in combination. Maybe I can live without that, but then you would need to remove from `stack.yaml` the pointer to `shell-stack.nix`. You also will have to update the instructions in the [README](https://github.com/ucsd-progsys/liquid-fixpoint/blob/deacde796cc666d7fbeec0fe23f1d587915401be/README.md#how-to-build-and-install).

There is a bigger question that is how to keep the flake setup working. It will require a flakes expert to maintain it, and if we add it to CI, I don't think we should make it a requirement to merge PRs, lest we bottleneck on the expert. Because of this difficulty, I think we need to make clear to users that flakes is not the recommended way to start with LF.

plredmond:
Thanks for your response @facundominguez. I've added two tasks to the PR description.

> There is a bigger question that is how to keep the flake setup working. It will require a flakes expert to maintain it, and if we add it to CI, I don't think we should make it a requirement to merge PRs, lest we bottleneck on the expert. Because of this difficulty, I think we need to make clear to users that flakes is not the recommended way to start with LF.

I use nixos full time, but I don't consider myself expert. I'd like it if this flake were simpler, but I don't currently know how. [Connor shared his rest-rewrite flake](https://github.com/ConnorBaker/rest/blob/master/flake.nix), from which I think I'll be able to simplify this one somewhat. That said, your point remains that updates may be nontrivial. Additionally, AFAIK the main developers of LH do not require nix, and so I agree that it doesn't make sense to block PRs on the flake.

ConnorBaker:
@plredmond Your flake actually inspired me to look at what I was doing again. I found I was able to get much of the same benefits of using `haskell.nix` with the existing `callCabal2nix` tools `nixpkgs` provides.

For example, one of the biggest benefits to me was that cabal and nix users would get the same version of packages (both used the `cabal.project.freeze` file). I realized that I could run `cabal freeze` inside `nix develop` and get a freeze file with the versions of the Hackage packages the flake’s version of `nixpkgs` provides (thus keeping the builds of cabal users consistent with those of the nix users).

After combing through `nixpkgs` standard library, I found [`updateManyAttrsByPath`](https://github.com/NixOS/nixpkgs/blob/master/lib/attrsets.nix#L119) which lets me do deep updates without nested `//` operators https://github.com/ConnorBaker/rest/blob/master/flake.nix#L74

facundominguez:
I'm merging this in #608.

plredmond:
As with https://github.com/ucsd-progsys/liquidhaskell/pull/1988 I think there are not yet completed changes in this PR. Should I make an issue for that? In either case, please assign to me any issues that arise from this merge.

facundominguez:
>  Should I make an issue for that?

Only if helpful to you.
We will learn soon enough if someone is relying on the shell.nix file that has been removed.

plredmond:
Yes, ok, we'll if that happens please assign to me.

PR #585, opened 2022-05-28
Follow hlint suggestion to avoid lambda.

philderbeast:
Another one for #552.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,5 +1,4 @@
 # Warnings currently triggered by your code
-- ignore: {name: "Avoid lambda"} # 4 hints
 - ignore: {name: "Eta reduce"} # 30 hints
 - ignore: {name: "Redundant bang pattern"} # 33 hints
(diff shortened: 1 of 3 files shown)


facundominguez:
Thanks @philderbeast. This is another one that we might want to continue to ignore, though. I find the lambda expressions easier to decipher than the point-free variants.

philderbeast:
In this instance the hlint suggestions cascaded, first avoid lambda then use second then redundant bracket and then use section or something like that.

I put a stop to it after the avoid lambda first suggestion with an hlint annotation on `subExprs'` to ignore the next suggestion, to use second.

facundominguez:
I'd rather not add HLint suggestions with the expectation that contributors will disable them where they make the code harder to read.

I'm closing this one without merging. But anyone feel free to challenge the choice.

facundominguez:
I'd rather not add HLint suggestions with the expectation that contributors will disable them where they make the code harder to read.

I'm closing this one. But anyone feel free to challenge the choice.

PR #590, opened 2022-05-29
Follow hlint suggestion, redundant bang pattern.

philderbeast:
I was wary about tackling this one for #552 as we don't have any memory, perf or benchmarking tests do we?

The docs back up the hlint suggestion.

> "Bang patterns do not have any effect with constructor patterns"
> SOURCE: ghc users guide on [bang patterns](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/strict.html#bang-patterns-informal)

Are we missing some bang patterns on strict fields of those constructor patterns?


facundominguez (approved):

facundominguez:
These look fine to me.

> we don't have any memory, perf or benchmarking tests do we?

We can run the LH tests and see that they don't take longer.

> Are we missing some bang patterns on strict fields of those constructor patterns?

`!(F a b)` should force the same as `(F a b)`, regardless of the strictness of the fields.

PR #593, opened 2022-06-01
Update REST to latest version

zgrannan:
The latest REST version incorporates a further optimization described here (https://github.com/zgrannan/rest/issues/9); the interface is changed slightly.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,5 +1,1 @@
 packages: .
-source-repository-package
-    type: git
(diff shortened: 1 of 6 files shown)


facundominguez:
Thanks @zgrannan! Are there any noticeable changes in the time it takes to verify LH tests?

zgrannan:
Nope, which isn't too surprising as most don't use REST. And for the ones that do, this change doesn't seem to make much of a difference.

PR #596, opened 2022-06-05
Follow hlint suggestion, eta reduce.

philderbeast:
The last of the batch for #552.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,4 +1,5 @@
 # Warnings currently triggered by your code
+- ignore: {name: "Eta reduce"} # 30 hints
 # The following are kept for usability and style sake
 - ignore: {name: "Avoid lambda"} # 4 hints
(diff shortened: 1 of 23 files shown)


facundominguez:
Thanks @philderbeast! After checking this PR, it also looks to me like we might want to ignore these warnings.

Sometimes the names of the eliminated variables are useless, and sometimes they convey some purpose which is helpful to read the code. I feel inclined to let users decide whether to eta reduce or not.

As always, please feel free to weigh in on the choice.

philderbeast:
Adding haddocks at the same time could make up for the loss of mostly one letter variables.

philderbeast:
I'll leave haddocks for another day (not going to help much anyway with module private functions) and suppress this hint so that we can be done with this batch of hlint related tasks.

facundominguez (approved):

PR #598, opened 2022-06-09
Use haskell/actions and bump hlint version.

philderbeast:
The `hlint-setup` and `hlint-run` actions have moved to the haskell organisation. Let's pick these actions up from there and bump the hlint version.


facundominguez (approved):
Thanks!

PR #599, opened 2022-06-09
Bump to ghc-9.2.3, bump actions, better caching.

philderbeast:
Got the caching snippets for cabal and stack from actions/cache. Put name before uses for each step.

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -20,6 +20,11 @@ jobs:
         with:
           submodules: true
 
+      - name: Setup z3
+        uses: pavpanchekha/setup-z3@1.2.2
+        with:
+          version: "4.8.7"
+
       - name: Setup GHC and cabal-install
         uses: haskell/actions/setup@v2
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/stack.yml
```diff
@@ -14,27 +14,37 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-stack-global-
+
+      - name: Cache .stack-work
+        uses: actions/cache@v3
+        with:
+          path: .stack-work
+          key: ${{ runner.os }}-stack-work-${{ hashFiles('stack.yaml') }}-${{ hashFiles('package.yaml') }}-${{ hashFiles('**/*.hs') }}
+          restore-keys: |
+            ${{ runner.os }}-stack-work-
```
facundominguez: Maybe
```suggestion
            ${{ runner.os }}-stack-work-${{ hashFiles('stack.yaml') }}
            ${{ runner.os }}-stack-work-
```
Otherwise, the retrieved cache may not match the needed dependencies.

Also, isn't adding `-${{ hashFiles('**/*.hs') }}` too specific? I doubt that we will get exact key matches if we keep that.

philderbeast: This I took as-is from https://github.com/actions/cache/blob/main/examples.md#haskell---stack. Thanks for checking it (I didn't). I'll tweak it for our use and to be more like the cabal caching.

.github/workflows/cabal.yml
```diff
@@ -13,41 +13,39 @@ jobs:
...
+      - name: Cache ~/.cabal/packages, ~/.cabal/store and dist-newstyle
+        uses: actions/cache@v3
         with:
-          version: "4.8.7"
+          path: |
+            ~/.cabal/packages
+            ~/.cabal/store
+            dist-newstyle
+          key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-${{ hashFiles('**/*.cabal', '**/cabal.project', '**/cabal.project.freeze') }}
+          restore-keys: ${{ runner.os }}-${{ matrix.ghc }}-cabal-
```
facundominguez: I was just noticing that perhaps this is going to cause caches to grow indefinitely as the dependencies change, right?

Would it be preferable to just drop `restore-keys` and start the cache afresh if there is not an exact key match?

philderbeast: We could drop `restore-keys` as it is only used when there's a cache miss, see [using the cache action](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action).

philderbeast: > I was just noticing that perhaps this is going to cause caches to grow indefinitely as the dependencies change, right?

> GitHub will remove any cache entries that have not been accessed in over 7 days. There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 10 GB.
> FROM: [usage limits and eviction policy](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy).

facundominguez:
Thanks @philderbeast!

PR #603, opened 2022-06-23
Ple with undecided guards

### LINKED ISSUE #602: Disable expansion of reflected functions with unevaluated guards
When working on #500, my plan was to stop generating equations in PLE like
```Haskell
length xs = if null xs then 0 else 1 + length (tail xs)
```
Instead, we would only produce `length xs = 0` or `length xs = 1 + length (tail xs)` when `xs` can be proven to be either empty or non-empty.

This would work most of the time, and many proofs are easy to adapt by splitting them in cases, but a couple of brainy modules in [liquidmeta](https://github.com/michaelborkowski/liquidmeta/) depended on the undecided guards to show up in equations anyway, and I ran out of steam to figure out fixing those.

PLE today generates `length xs = if null xs then 0 else 1 + length (tail xs)` at the end of the algorithm, for every application of `length` in candidate expressions whose guards couldn't be decided. Because most of the time these equations aren't needed, I'd like to propose putting them behind a flag. So the `--ple` behavior would omit these equations with undecided guards by default, but if the user wants them, they can be generated with a flag `--ple-with-undecided-guards`.

The purpose of this is to make PLE effect on the SMT solver more predictable, and to clean the set of generated equations. I don't anticipate any performance improvements because of this.

How does that sound?

facundominguez:
Closes #602

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,8 +44,8 @@ jobs:
           name: Test
           command: |
             mkdir -p /tmp/junit
-            stack --no-terminal test -j2 liquid-fixpoint:test --flag liquid-fixpoint:devel --test-arguments="--xml=/tmp/junit/main-test-results.xml"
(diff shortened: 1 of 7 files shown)


nikivazou:
@ranjitjhala you will merge this, correct?

facundominguez:
Merged this after merging of ucsd-progsys/liquidhaskell#2016

PR #605, opened 2022-07-04
Only unfold if no undecided guard remains in an equation

facundominguez:
A couple of tests in LH need to be changed as a result of this: facundominguez/liquidhaskell@814b84d

If we have functions like

    reflect at
    at :: xs:[a] -> { i:Int | 0<= i  && i < length xs } -> [a]
    at (y:_ys) 0 = y
    at (_y:ys) i = at ys (i-1)

    reflect filter
    filter :: (a -> Bool) -> xs:[a] -> [a]
    filter p [] = []
    filter p (y:ys) = if p y then y : filter p ys else filter p ys

Before this PR, equations are unfolded as soon as the first guard is decided.
* In the case of `at`, we don't need to know whether the index is `0`.
* In the case of `filter`, we don't need to know what is the return value of `p y`.

After this PR, we don't unfold unless all the guards are decided.
* In the case of `at`, we need to know that the list is non-empty and whether the index is `0`.
* In the case of `filter`, we need to know if the list is non-empty and what is the return value of `p y`.

### DIFF
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -879,10 +879,9 @@ shortcut :: Knowledge -> ICtx -> EvalType -> Expr -> [Expr] -> EvalST (Expr, Boo
 shortcut γ ctx et (EIte i e1 e2) es2 = do
       (b, _) <- eval γ ctx et i
       b'  <- mytracepp ("evalEIt POS " ++ showpp (i, b)) <$> isValidCached γ b
-      let changed (a, _, c) = (a, True, c)
       case b' of
-        Just True -> changed <$> shortcut γ ctx et e1 es2
-        Just False -> changed <$> shortcut γ ctx et e2 es2
+        Just True -> shortcut γ ctx et e1 es2
+        Just False -> shortcut γ ctx et e2 es2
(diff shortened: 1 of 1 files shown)


facundominguez:
I'm torn on this one.

The equations produced by PLE are smaller if we delay unfolding until we have more information, but the user needs to have some more awareness of how PLE works. Sometimes she might want PLE to unfold anyway, and this will require explicit use of `pleUnfold` as shown in the tests referred in the description.

On the plus side:
* `pleUnfold` seems to be needed still in few tests.
* The user is made aware of how the SMT solver verifies a constraint. e.g.
  * Does it need only facts in the environment?
  * Does it need only to unfold some definitions?
  * Does it need to reason about if-statements with an undecided condition?

I think these provide the user with needed insight on how to fix the verification when the program is changed.

On the cons side, verification is less magical/automated, which means she needs to invest some more upfront effort in understanding PLE before using it effectively in some cases. Probably this is the case already in other areas of LH.

ranjitjhala:
I think it’s ok — my general syntactic heuristic is the “proof” over a
function needs to generally “unfold”/split cases as much as the function
itself - Ie generally mirrors the guard structure of the function… and this
restriction (don’t unfold under guards) more faithfully implements that
heuristic imo ..

PR #607, opened 2022-07-07
Have gh-action file name match action name.

philderbeast:
This is what we do for the other two gh-actions.


facundominguez (approved):
LGTM!

ranjitjhala:
Awesome, thank you!

Issue #610, opened 2022-07-30
Stan static analysis using `.hie` files.


philderbeast:
I've tried running [stan](https://github.com/kowainik/stan) over liquid-fixpoint. Some of its suggestions I expect we'd want to pursue. Like with the recent hlint changes there are some hints we'd ignore and there is a config file where we can turn those off.

philderbeast:
For the foldl anti-pattern stan finds a few hints like the following:

![Screen Shot 2022-07-30 at 9 23 58 AM](https://user-images.githubusercontent.com/633283/181916414-daabc682-5b85-4874-9b6f-a79460e27579.png)

The same thing from the command line:

```
  File:         src/Language/Fixpoint/Parse.hs
  Module:       Language.Fixpoint.Parse
  LoC:          1652
  Observations: 1
  Extensions from .cabal:
  Extensions from module: FlexibleContexts, FlexibleInstances, NoMonomorphismRestriction, UndecidableInstances, DeriveGeneric, OverloadedStrings
 ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 ┃  ✦ ID:            OBS-STAN-0202-6VNRVE-1385:18
 ┃  ✦ Severity:      Error
 ┃  ✦ Inspection ID: STAN-0202
 ┃  ✦ Name:          Anti-pattern: foldl
 ┃  ✦ Description:   Usage of space-leaking function 'foldl'
 ┃  ✦ Category:      #SpaceLeak #AntiPattern
 ┃  ✦ File:          src/Language/Fixpoint/Parse.hs
 ┃
 ┃  1384 ┃
 ┃  1385 ┃     rwMap      = foldl insert (M.fromList []) rwEntries
 ┃  1386 ┃                  ^^^^^
 ┃
 ┃  💡 Possible solution:
 ┃      ⍟ Replace 'foldl' with 'foldl''
 ┃      ⍟ Use 'foldr (flip . f)` instead of 'foldl f'
```

PR #611, opened 2022-07-30
Add stan gh-action.

philderbeast:
Adds a stan GitHub action with a configuration that turns off all hints. Like we did for hlint recently, we can fix some of these one by one and choose to ignore others, see #610.


facundominguez (approved):
Thanks @philderbeast!

I think it is worth trying it for a while.

PR #613, opened 2022-08-02
Handles are stdin, stdout, and stderr respectively.

philderbeast:
I was trying out `typed-process` on this repo as a learning exercise and noticed that the call to `runInteractiveCommand` has flipped the names of the handles.

> `runInteractiveCommand :: String -> IO (Handle, Handle, Handle, ProcessHandle)`
> Runs a command using the shell, and returns Handles that may be used to communicate with the process via its stdin, stdout, and stderr respectively.
> [runInteractiveCommand](https://hackage.haskell.org/package/process-1.6.14.0/docs/System-Process.html#v:runInteractiveCommand)

### DIFF
--- a/src/Language/Fixpoint/Smt/Interface.hs
+++ b/src/Language/Fixpoint/Smt/Interface.hs
@@ -264,7 +264,7 @@ smtWriteRaw me !s = {- SCC "smtWriteRaw" -} do
 
 -- | Reads a line of output from the SMT solver.
 smtReadRaw :: Context -> IO T.Text
-smtReadRaw me = {- SCC "smtReadRaw" -} TIO.hGetLine (ctxOut me)
+smtReadRaw me = TIO.hGetLine (ctxOut me)
 {-# SCC smtReadRaw  #-}
 
 hPutStrLnNow :: Handle -> LT.Text -> IO ()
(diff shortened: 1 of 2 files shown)


src/Language/Fixpoint/Smt/Interface.hs
```diff
@@ -254,18 +254,20 @@ negativeP
...
+smtWriteRaw :: Context -> Raw -> IO ()
 smtWriteRaw me !s = {- SCC "smtWriteRaw" -} do
   -- whenLoud $ do LTIO.appendFile debugFile (s <> "\n")
   --               LTIO.putStrLn ("CMD-RAW:" <> s <> ":CMD-RAW:DONE")
-  hPutStrLnNow (ctxCout me) s
+  hPutStrLnNow (ctxIn me) s
   maybe (return ()) (`LTIO.hPutStrLn` s) (ctxLog me)
 
-smtReadRaw       :: Context -> IO T.Text
-smtReadRaw me    = {- SCC "smtReadRaw" -} TIO.hGetLine (ctxCin me)
```
philderbeast, line 265: Can I remove `{- SCC "smtReadRaw" -}`? It is there as a pragma on the next line.

facundominguez: I think so.

facundominguez (approved):
LGTM!

PR #615, opened 2022-08-03
Use foldl' instead of foldl.

philderbeast:
Reported by stan, see https://wiki.haskell.org/Foldr_Foldl_Foldl'.


facundominguez (approved):
LGTM!

PR #616, opened 2022-08-03
Missed changing the link when renaming hlint-ci.

philderbeast:
Before the fix:

Screen Shot 2022-08-02 at 9 56 59 PM


facundominguez (approved):
LGTM!

PR #619, opened 2022-08-05
Follow stan hint, explicit fixity (default is infixl 9).

philderbeast:
See #610. I'm putting this up here as-is so please label with WIP. I'd like some feedback on which fixities to go with. Are there some that shouldn't be associative? Could some of these be replaced with existing operators we pull in from elsewhere?

https://kowainik.github.io/posts/fixity#standard-examples

For now I've just been explicit but with the default `infixl 9`.

### DIFF
--- a/src/Language/Fixpoint/Misc.hs
+++ b/src/Language/Fixpoint/Misc.hs
@@ -5,7 +5,6 @@
 {-# LANGUAGE TupleSections             #-}
 {-# LANGUAGE ConstraintKinds           #-}
 {-# LANGUAGE TypeOperators             #-}
-{-# LANGUAGE BangPatterns              #-}
 {-# LANGUAGE ImplicitParams            #-} -- ignore hlint
 
 
@@ -308,16 +307,6 @@ writeLoud s = whenLoud $ putStrLn s >> hFlush stdout
 ensurePath :: FilePath -> IO ()
 ensurePath = createDirectoryIfMissing True . takeDirectory
 
-singleton :: a -> [a]
-singleton x = [x]
-
-pair :: a -> a -> [a]
-pair x1 x2 = [x1, x2]
-
-triple :: a -> a -> a -> [a]
-triple x1 x2 x3 = [x1, x2, x3]
-
-
 fM :: (Monad m) => (a -> b) -> a -> m b
 fM f = return . f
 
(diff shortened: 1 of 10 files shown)


src/Language/Fixpoint/Misc.hs
```diff
@@ -425,12 +425,15 @@ allCombinations xs = assert (all ((length xs == ) . length)) $ go xs
 powerset :: [a] -> [[a]]
 powerset xs = filterM (const [False, True]) xs
 
+infixl 9 =>>
 (=>>) :: Monad m => m b -> (b -> m a) -> m b
 (=>>) m f = m >>= (\x -> f x >> return x)
 
+infixl 9 <<=
 (<<=) :: Monad m => (b -> m a) -> m b -> m b
 (<<=) = flip (=>>)
```
facundominguez, line 423: I would think these should be removed from here. And replaced in the code with `>>=` and `=<<`.

src/Language/Fixpoint/Misc.hs
```diff
@@ -425,12 +425,15 @@ allCombinations xs = assert (all ((length xs == ) . length)) $ go xs
...
 
+infixl 9 <$$>
 (<$$>) ::  (Monad m) => (a -> m b) -> [a] -> m [b]
 _ <$$> []           = return []
 f <$$> [x1]         = singleton <$> f x1
```
facundominguez: Maybe it is worth replacing this with `traverse` in the code.

philderbeast:
Looks like `(=>>) :: Monad m => m b -> (b -> m a) -> m b` and `(<<=)` are not used.

src/Language/Fixpoint/Misc.hs
philderbeast: Looks like `(=>>) :: Monad m => m b -> (b -> m a) -> m b` and `(<<=)` are not used.

philderbeast: Had a quick peek at `liquidhaskell`. Each operator is only used once there, `(=>>)` in `Language.Haskell.Liquid.Constraint.Fresh` and `(<<=)` in `Language.Haskell.Liquid.Constraint.Generate`. Perhaps those modules should define such an operator privately or use the inline equivalent and we remove them from `liquid-fixpoint`?

facundominguez (approved):
LGTM!

Issue #614, opened 2022-08-02
`z3-4.10.*` versions don't work for liquid-fixpoint


philderbeast:
After `brew upgrade z3` I found that `cabal test all` for `liquid-fixpoint` had failures.

Incidentally I found that `pavpanchekha/setup-z3@1.2.2` can't install the latest version of z3 as ubuntu binaries are no longer being built or maybe they are under another name (a separate issue).

```yaml
      - name: Setup z3
        uses: pavpanchekha/setup-z3@1.2.2
        with:
          version: "4.8.7"
```

This `pavpanchekha/setup-z3` file name construction doesn't work for some of the newer releases.

```js
    let path = "https://github.com/Z3Prover/z3/releases/download/z3-" + version;
    let file = "z3-" + version + "-" + architecture + "-" + distribution + ".zip";
```

![Screen Shot 2022-08-02 at 7 16 50 AM](https://user-images.githubusercontent.com/633283/182362157-94bfd46e-c5e3-4831-8114-cdd4b399773e.png)

I uninstalled with `brew uninstall z3`, downloaded binaries for osx to `$HOME/dev/tools/z3/` and used [direnv](https://direnv.net/) in the `liquid-fixpoint` project root to switch z3 versions that I tested with `cabal test all`.

```
# .envrc file contents

# These z3 versions work for liquid-fixpoint's test-suite.
# PATH_add "$HOME/dev/tools/z3/z3-4.8.7-x64-osx-10.14.6/bin"
# PATH_add "$HOME/dev/tools/z3/z3-4.8.10-x64-osx-10.15.7/bin"
# PATH_add "$HOME/dev/tools/z3/z3-4.8.17-x64-osx-10.16/bin"
# PATH_add "$HOME/dev/tools/z3/z3-4.9.0-x64-osx-10.16/bin"

# The z3-4.10.* versions don't work for liquid-fixpoint.
PATH_add "$HOME/dev/tools/z3/z3-4.10.0-x64-osx-10.16/bin"
# PATH_add "$HOME/dev/tools/z3/z3-4.10.2-x64-osx-10.16/bin"
```

philderbeast:
I've raised an issue with `setup-z3` for download file names, https://github.com/pavpanchekha/setup-z3/issues/7

ranjitjhala:
wow thanks for catching this!

philderbeast:
Trying the example from the readme:

```
> z3 --version
[...]

PR #621, opened 2022-08-06
Use setup-z3 for `z3-4.8.7` and `z3-4.10.2`

philderbeast:
Related to #614. See also https://github.com/pavpanchekha/setup-z3/issues/7. I'm only proposing a matrix of z3 for the cabal action. Other actions get `z3-4.10.2`.

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build:
-    name: ghc ${{ matrix.ghc }}
+    name: ghc-${{ matrix.ghc }} z3-${{ matrix.z3 }}
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


facundominguez:
Thanks @philderbeast. Looks good to me. I'd say it isn't worth testing more than one version of z3. Though I'm fine accepting this as is, if someone really cares.

philderbeast:
> I'd say it isn't worth testing more than one version of z3.

I was using this to help test https://github.com/pavpanchekha/setup-z3/issues/7. There's a lot of releases of z3 between `z3-4.8.7` and `z3-4.10.2` so I thought it prudent to try both in `liquid-fixpoint` at least for a short while.

philderbeast:
Yes @facundominguez let's go with the latest z3 version.

If we get a failure, comparing with an older version locally is going to be way easier than maintaining conditionals in a gh-action.

facundominguez (approved):

PR #624, opened 2022-10-05
Add mapToSet and mapPrj

andrewthad:
Adds `mapToSet` and `mapPrj`, defined like this:

```
(define-fun smt_map_prj ((s LSet) (m Map)) Map ((_ map (ite (Bool Elt Elt) Elt)) s m ((as const (Array Elt Elt)) 0)))
(define-fun smt_map_to_set ((m Map)) LSet ((_ map (> (Elt Elt) Bool)) m ((as const (Array Elt Elt)) 0)))
```

The intent is to make it possible (in the liquidhaskell repo) to write a function that unions two bags (in `Language.Haskell.Liquid.Bag`) by choosing the first non-zero element rather than by summing the elements.

The implementations in this PR bake in zero as a special value, further committing the `Map` type in liquid-fixpoint to bag interpretation.

### DIFF
--- a/src/Language/Fixpoint/Smt/Theories.hs
+++ b/src/Language/Fixpoint/Smt/Theories.hs
@@ -71,7 +71,8 @@ elt  = "Elt"
 set  = "LSet"
 map  = "Map"
 
-emp, sng, add, cup, cap, mem, dif, sub, com, sel, sto, mcup, mdef, mprj, mToSet :: Raw
+emp, sng, add, cup, cap, mem, dif, sub, com, sel, sto, mcup, mdef, mprj :: Raw
+mToSet, mshift, mmax, mmin :: Raw
 emp   = "smt_set_emp"
 sng   = "smt_set_sng"
 add   = "smt_set_add"
@@ -84,12 +85,16 @@ com   = "smt_set_com"
 sel   = "smt_map_sel"
 sto   = "smt_map_sto"
 mcup  = "smt_map_cup"
+mmax  = "smt_map_max"
+mmin  = "smt_map_min"
 mdef  = "smt_map_def"
 mprj  = "smt_map_prj"
+mshift = "smt_map_shift"
 mToSet = "smt_map_to_set"
 
 
-setEmpty, setEmp, setCap, setSub, setAdd, setMem, setCom, setCup, setDif, setSng :: Symbol
+setEmpty, setEmp, setCap, setSub, setAdd, setMem, setCom, setCup :: Symbol
+setDif, setSng :: Symbol
 setEmpty = "Set_empty"
 setEmp   = "Set_emp"
 setCap   = "Set_cap"
@@ -102,11 +107,15 @@ setDif   = "Set_dif"
 setSng   = "Set_sng"
 
 mapSel, mapSto, mapCup, mapDef, mapPrj, mapToSet :: Symbol
+mapMax, mapMin, mapShift :: Symbol
 mapSel   = "Map_select"
 mapSto   = "Map_store"
 mapCup   = "Map_union"
+mapMax   = "Map_union_max"
+mapMin   = "Map_union_min"
 mapDef   = "Map_default"
 mapPrj   = "Map_project"
+mapShift = "Map_shift" -- See [Map key shift]
 mapToSet = "Map_to_set"
 
 -- [Interaction between Map and Set]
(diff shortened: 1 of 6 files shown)


ranjitjhala:
Nice! Can you add a test (fq) or smt2 that exercises this functionality?

andrewthad:
I've added some fq files and run the new tests by hand. When I try to run all the tests with:

```
cabal build test
./dist-newstyle/build/x86_64-linux/ghc-8.10.7/liquid-fixpoint-0.8.10.7.1/t/test/build/test/test
```

I get a `ExitFailure 127` for everything, but it looks like it's because `fixpoint` is not on my path. I'm not sure why the test suite would need the binary since it should just have access to the library directly. I doubt that there are any regressions to existing tests though.

andrewthad:
I've got one more weird one to throw in here. Don't merge this yet.

src/Language/Fixpoint/Smt/Theories.hs
```diff
@@ -99,11 +101,31 @@ setCup   = "Set_cup"
...
+-- [Interaction between Map and Set]
+-- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+-- Function mapToSet: Convert a map to a set. The map's key may be of
+-- any type and is preserved as the set's element type. More precisely:
+--   mapToSet : Map k Int -> Set k
+-- The element type must be Int. All non-positive elements are mapped
+-- to False, and all positive elements are mapped to True. In practice,
+-- negative elements should not exist because Map is intended to be used
+-- as a bag, so the element is a non-negative number representing
+-- the occurrences of its corresponding key.
```
facundominguez, line 130: There seems to be some assumptions implicit here that I don't have. For instance: How is the value of the integers in the map relevant to the conversion function?

I'm discovering that I can't find where the sorts Set and Map or their operations are defined.

andrewthad: `Map` is `Array Int Int`, and `Set` is `Array Int Bool`. There's a big prelude at the top of many of the smt files generated by liquid-fixpoint that looks like this:

```
(define-sort Elt () Int)
(define-sort LSet () (Array Elt Bool))
(define-fun smt_set_emp () LSet ((as const LSet) false))
(define-fun smt_set_sng ((x Elt)) LSet (store ((as const LSet) false) x true))
(define-fun smt_set_mem ((x Elt) (s LSet)) Bool (select s x))
(define-fun smt_set_add ((s LSet) (x Elt)) LSet (store s x true))
(define-fun smt_set_cup ((s1 LSet) (s2 LSet)) LSet ((_ map or) s1 s2))
(define-fun smt_set_cap ((s1 LSet) (s2 LSet)) LSet ((_ map and) s1 s2))
(define-fun smt_set_com ((s LSet)) LSet ((_ map not) s))
(define-fun smt_set_dif ((s1 LSet) (s2 LSet)) LSet (smt_set_cap s1 (smt_set_com s2)))
(define-fun smt_set_sub ((s1 LSet) (s2 LSet)) Bool (= smt_set_emp (smt_set_dif s1 s2)))
(define-sort Map () (Array Elt Elt))
(define-fun smt_map_sel ((m Map) (k Elt)) Elt (select m k))
(define-fun smt_map_sto ((m Map) (k Elt) (v Elt)) Map (store m k v))
(define-fun smt_map_cup ((m1 Map) (m2 Map)) Map ((_ map (+ (Elt Elt) Elt)) m1 m2))
(define-fun smt_map_prj ((s LSet) (m Map)) Map ((_ map (ite (Bool Elt Elt) Elt)) s m ((as const (Array Elt Elt)) 0)))
(define-fun smt_map_to_set ((m Map)) LSet ((_ map (> (Elt Elt) Bool)) m ((as const (Array Elt Elt)) 0)))
(define-fun smt_map_max ((m1 Map) (m2 Map)) Map (lambda ((i Int)) (ite (> (select m1 i) (select m2 i)) (select m1 i) (select m2 i))))
(define-fun smt_map_min ((m1 Map) (m2 Map)) Map (lambda ((i Int)) (ite (< (select m1 i) (select m2 i)) (select m1 i) (select m2 i))))
(define-fun smt_map_shift ((n Int) (m Map)) Map (lambda ((i Int)) (select m (- i n))))
(define-fun smt_map_def ((v Elt)) Map ((as const (Map)) v))
```

All of those are defined in `z3Preamble`.

> How is the value of the integers in the map relevant to the conversion function?

If the conversion function exists at all, then the values *must* be relevant. The `Array` type from smtlib maps all keys to a value, so there are no positions at which a value is absent. I chose to interpret `Map` more specifically to mean `Bag` for this function because that seems to be its only use in liquid haskell's standard library.

andrewthad:
I've add three more, with tests, in the latest commit. One unions maps, combining elements with `max`. Another unions maps, combining elements with `min`. And the last one shifts maps.

Also, I can just remove `smt_map_to_set` if it's objectionable. Now that I've figured out how to use Z3's `lambda` to build the other combinators, I don't need it anymore.

nikivazou:
Hey @andrewthad do you really need the `smt_map_shift` operation? It is implemented using the lambda syntax and it is not build in on cvc5, so unless there is a real need for it, maybe we can remove it or make it "optional". 

Or ideally have it user-defined in some way instead of build it.

andrewthad:
I don't need any of these anymore. I had been trying to see how far I could get with having Liquid Haskell enforce that contexts (for a lambda calculus implementation) were well formed. It actually worked pretty well, but I've stopped pursuing this approach. So if you need to get rid of any of these to make cvc5 work, get rid of them.

nikivazou:
Thanks!!!

Issue #622, opened 2022-08-27
Produce an intelligible error message when the z3 version is not the expected one


facundominguez:
Our CI is using z3 as SMT solver, and new users usually try that when first installing LH. However, LF doesn't work with any z3 version, leading to errors like
```
crash: SMTLIB2 respSat = Error "line 4 column 27: unknown parameter 'model_partial', this is an old parameter name, invoke 'z3 -p' to obtain the new parameter list"
```

It would help users if LF checked the z3 version being used and produced an error if the version is not one of the expected ones.

At this point I don't know which z3 versions are supported, but we could start by demanding the specific version used in CI, and maybe adding a flag to disable the check if the user wants to run LF with whatever version she has instead.

ranjitjhala:
Yes this would be good — that error (the person reporting on `slack`?) surprised me — as LH has been mostly very stable since we use the SMTLIB interface which hasn’t changed in awhile.

PR #626, opened 2022-10-20
Support the newest Z3 version (z3-4.11.2)

Alf0nso:
Running Liquid Haskell with newer versions of z3 causes this error:
```
crash: SMTLIB2 respSat = Error "line 4 column 27: unknown parameter 'model_partial', this is an old parameter name, invoke 'z3 -p' to obtain the new parameter list"
```
Mentioned in this issue: https://github.com/ucsd-progsys/liquid-fixpoint/issues/622
In this pr we removed the z3 "model.partial false".
On older versions of z3 the default option was already false.
Invoke `z3 -p` and we get:
```
[module] model
    ...
    partial (bool) (default: false)
```

### DIFF
--- a/src/Language/Fixpoint/Smt/Interface.hs
+++ b/src/Language/Fixpoint/Smt/Interface.hs
@@ -523,8 +523,7 @@ makeMbqi cfg
 z3_options :: [LT.Text]
 z3_options
   = [ "(set-option :auto-config false)"
-    , "(set-option :model true)"
-    , "(set-option :model.partial false)"]
+    , "(set-option :model true)" ]
 
 
 


nikivazou:
Thanks @Alf0nso!!!!!! Can we have a pull request in LH too that links to this change?

oquechy:
> Thanks @Alf0nso!!!!!! Can we have a pull request in LH too that links to this change?

@nikivazou do you think we can do this before merging the new change into fixpoint?

nikivazou:
yes, lets make sure nothing breaks in LH before merging this in fixpoint!

PR #628, opened 2022-11-03
Fix z3 version checks

facundominguez:
We were comparing 4.11.2 with 4.3.2 by comparing
["4", "11", "2"] >= ["4", "3", "2"]

Which would return False because "11" >= "3" is false.

This patch changes this to read the version components as integers instead of strings.

Tested with z3-4.11.0.


ranjitjhala:
Wow thank you!!!

PR #629, opened 2022-11-11
remove nix flake

plredmond:
I'm no longer able to keep up maintaining this flake for my personal use. I suggest removing it, because it's very complex, and most users are probably satisfied with cabal and stack.


ranjitjhala:
ok! I can remove it! was it in the CI?

plredmond:
I'm fairly certain that it's not used by CI. We might confirm that with @facundominguez.

facundominguez (approved):
LGTM

PR #630, opened 2022-11-15
Show ($k x y) such that it can be fed back to the parser

shingarov:
Even though Haskell is not, by design, a homoiconic language a-la Lisp, in some cases having Show produce strings that can be parsed back is advantageous.  For example, LiquidFixpoint's clients such as Sprite, may dump VCs in LF format; these dumps can be a valuable aid in understanding the code.  This series of commits ensures the dumped VCs are in fact valid LF.


ranjitjhala:
Thank you so much!

PR #631, opened 2022-11-15
Allow AND of zero conjuncts

shingarov:
Cf. PR630


ranjitjhala:
Thanks again @shingarov !!!

PR #641, opened 2023-01-23
`Language.Fixpoint.Smt.Interface`: use `smtlib-backends` for managing the solver process

qaristote:
This PR replaces the code for creating, writing to, reading from, and closing the SMT solver external process by calls to [the smtlib-backends library](https://github.com/tweag/smtlib-backends), "a Haskell library providing low-level functions for SMTLIB-based interaction with SMT solvers". There are two main advantages to using this library:
- there is less code to maintain on the side of liquid-fixpoint;
- additionally to running solvers as external processes, smtlib-backends also offers the option to send SMTLIB commands to Z3 through its bindings, which is both faster and less error-prone.

The code is definitely hacky as liquid-fixpoint wasn't designed around the use of smtlib-backends and I just tried to get a working version, so the point of this PR is not to be merged as is (although this is still possible) but rather to create a discussion around the use of smtlib-backends.

# What changed

## In `Language.Fixpoint.Types.Config`

The `SMTSolver` datatype is augmented with the `Z3mem` constructor which represents communication with Z3 through its bindings. Since this is faster and safer than running solvers as external processes, this constructor is set as the default value for this datatype.

## In `Language.Fixpoint.Smt.Types`

The `Context` datatype no longer contains a PID and input and output channels, but instead a `SMTLIB.Backends.Solver`, a high-level wrapper to SMT solver backends represented by the new `ContextHandle` datatype. This latter datatype is just a sum type of the different backend types available in smtlib-backends, that is `SMTLIB.Backends.Process.Handle` datatype from smtlib-backends-process (for running a solver as an external process) and the `SMTLIB.Backends.Z3.Handle` datatype from smtlib-backends-z3 (for communicating with Z3 through its bindings).

The `Context` datatype also no longer contains the fields used when sending commands asynchronously: I decided to not re-implement this for now, but it's definitely possible if deemed necessary.

## In `Language.Fixpoint.Smt.Interface`

Instead of manually spawing an external process, the `makeContext` function now uses the `new` function from either `SMTLIB.Backends.Process` or `SMTLIB.Backends.Z3` for initializing the solver backend. The corresponding `ContextHandle` is kept inside the context so that they're available at cleaning-up time.

The `SMTLIB.Backends.Solver` object is created from this backend along with the `Queuing` parameter. The latter means the solver comes with a queue to write commands to. This queue is only flushed and actually sent to the solver when absolutely necessary (that is when we need the solver's response), reducing the amount of interaction between liquid-fixpoint and the solver backend (sending commands to the solver is the costlier part of the interaction hence we try to do it as rarely as possible).

The writing of commands and reading of responses on the solver's input and output channels are replaced with calls to the `SMTLIB.Backends.command` and `SMTLIB.Backends.command_` functions. Since these functions write the command and read the solver's response in one atomic operation, this required adding an `IORef ByteString` buffer for holding the solver's response until liquid-fixpoint actually tries to read it.

The body of the `asyncCommand` function was replaced by a call to `smtWrite`, and the parallel process that sends the commands asynchronously was removed.

# Benchmarks

Here are the results I got when running LiquidHaskell's benchmarks:
perf

For most of the benchmarks, it seems the use of smtlib-backends doesn't matter much as LiquidHaskell probably doesn't spend a lot of time interacting with the SMT solver overall. But there are a few other benchmarks where using smtlib-backends seems to yield a significant speed-up.

Since this PR removes sending commands asynchronously, I also tried comparing my branch to [a version of liquid-fixpoint where I removed asynchronicity](https://github.com/ucsd-progsys/liquid-fixpoint/compare/develop...qaristote:liquid-fixpoint:qa/no_async):
perf

The results look weird to me as using smtlib-backends introduces both huge speed-ups and huge slow-downs. But I actually get the same differences when comparing liquid-fixpoint with no asynchronicity vs plain liquid-fixpoint, so this doesn't look like a problem with my branch:
perf

(I might have made some kind of mistake when running these benchmarks though).

# What's next

Ideally, liquid-fixpoint would be redesigned a bit around the use of smtlib-backends before it is actually merged. This would mainly entail not separating sending commands from reading the solver's response, since this requires the kinda dirty hack of keeping the solver's response around in a buffer. I'm not sure the intermediate representation of commands as `Text` is necessary anymore (smtlib-backends uses bytestrings).

### DIFF
--- a/src/Language/Fixpoint/Smt/Interface.hs
+++ b/src/Language/Fixpoint/Smt/Interface.hs
@@ -273,7 +273,7 @@ makeProcess
   -> Process.Config
   -> IO (SMTLIB.Backends.Backend, IO ())
 makeProcess ctxLog cfg
-  = do handle@Process.Handle {..} <- Process.new cfg
+  = do handle@Process.Handle {hMaybeErr = Just hErr, ..} <- Process.new cfg
        case ctxLog of
          Nothing -> return ()
          Just hLog -> void $ async $ forever
@@ -292,11 +292,17 @@ makeContext' cfg ctxLog
            {- "z3 -smt2 -in"                   -}
            {- "z3 -smtc SOFT_TIMEOUT=1000 -in" -}
            {- "z3 -smtc -in MBQI=false"        -}
-           makeProcess ctxLog $ Process.Config "z3" ["-smt2", "-in"]
+           makeProcess ctxLog $ Process.defaultConfig
+                             { Process.exe = "z3"
+                             , Process.args = ["-smt2", "-in"] }
          Z3mem   -> Conditional.Z3.makeZ3
-         Mathsat -> makeProcess ctxLog $ Process.Config "mathsat" ["-input=smt2"]
+         Mathsat -> makeProcess ctxLog $ Process.defaultConfig
+                             { Process.exe = "mathsat"
+                             , Process.args = ["-input=smt2"] }
          Cvc4    -> makeProcess ctxLog $
-                      Process.Config "cvc4" ["--incremental", "-L", "smtlib2"]
+                      Process.defaultConfig
+                             { Process.exe = "cvc4"
+                             , Process.args = ["--incremental", "-L", "smtlib2"] }
        solver <- SMTLIB.Backends.initSolver SMTLIB.Backends.Queuing backend
        loud <- isLoud
        return Ctx { ctxSolver  = solver
--- a/src/Language/Fixpoint/Smt/Interface.hs
+++ b/src/Language/Fixpoint/Smt/Interface.hs
@@ -273,14 +273,13 @@ makeProcess
   -> Process.Config
   -> IO (SMTLIB.Backends.Backend, IO ())
 makeProcess ctxLog cfg
-  = do handle@Process.Handle {..} <- Process.new $ cfg
+  = do handle@Process.Handle {..} <- Process.new cfg
        case ctxLog of
          Nothing -> return ()
-         Just hLog -> void $ async $
-           (forever $ do
-               err <- LTIO.hGetLine $ hErr
+         Just hLog -> void $ async $ forever
+           (do err <- LTIO.hGetLine hErr
                LTIO.hPutStrLn hLog $ "OOPS, SMT solver error:" <> err
-           ) `catch` \SomeException {} -> return ()
+           ) `catch` \ SomeException {} -> return ()
        let backend = Process.toBackend handle
        hSetBuffering hOut $ BlockBuffering $ Just $ 1024 * 1024 * 64
        hSetBuffering hIn $ BlockBuffering $ Just $ 1024 * 1024 * 64
--- a/src/Language/Fixpoint/Smt/Interface.hs
+++ b/src/Language/Fixpoint/Smt/Interface.hs
@@ -93,7 +93,6 @@ import           System.Directory
 import           System.Console.CmdArgs.Verbosity
 import           System.FilePath
 import           System.IO
-import qualified System.Process.Typed
 import qualified Data.Attoparsec.Text     as A
 -- import qualified Data.HashMap.Strict      as M
 import           Data.Attoparsec.Internal.Types (Parser)
@@ -105,6 +104,7 @@ import           Language.Fixpoint.Utils.Builder as Builder
 import qualified SMTLIB.Backends
 import qualified SMTLIB.Backends.Process as Process
 import qualified Language.Fixpoint.Conditional.Z3 as Conditional.Z3
+import Control.Concurrent.Async (async)
 
 {-
 runFile f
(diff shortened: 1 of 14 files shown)


src/Language/Fixpoint/Smt/Interface.hs
```diff
@@ -255,41 +267,47 @@ negativeP
...
 smtReadRaw me = do
-  eof <- hIsEOF (ctxOut me)
-  if eof then Misc.errorstar "SMT returned End of File" else
-    TIO.hGetLine (ctxOut me)
-{-# SCC smtReadRaw  #-}
-
-hPutStrLnNow :: Handle -> LT.Text -> IO ()
-hPutStrLnNow h !s = LTIO.hPutStrLn h s >> hFlush h
-{-# SCC hPutStrLnNow #-}
+  respLn <- atomicModifyIORef (ctxResp me) $ \resps ->
```
facundominguez: What is the need to call `atomicModifyIORef` here?

facundominguez:
Thanks @qaristote! I'm guessing the benchmark measuring needs to be sharpened a bit still.

src/Language/Fixpoint/Smt/Interface.hs
```diff
@@ -85,9 +86,6 @@ module Language.Fixpoint.Smt.Interface (
 
     ) where
 
-import           Control.Concurrent.Async (async, cancel)
-import           Control.Concurrent.STM
-  (TVar, atomically, modifyTVar, newTVarIO, readTVar, retry, writeTVar)
```
facundominguez, line 90: I'd prefer we stick with STM for concurrency.

ranjitjhala:
i’m very excited by this PR thank you!

src/Language/Fixpoint/Smt/Interface.hs
qaristote: None, I probably wrote this at a time when I wasn't sure of the concurrency model of liquid-fixpoint yet.

qaristote: Removed in https://github.com/ucsd-progsys/liquid-fixpoint/pull/641/commits/cb9ff514fbcbe3f1c8ec08997391620d0871feb5.

facundominguez:
I squashed the commits from @qaristote laying in some other branch, and made linking with Z3 conditionally. `liquid-fixpoint` only links Z3 if requested with a cabal flag.

One other change with respect to the previous state of the PR is that the functions to interact asynchronously with the SMT solver have been removed. When measuring them we didn't find any improvements, and they made the code more complicated.

There are some test failures to check in the CI jobs still.

In the job "cabal / ghc-9.2.3 z3-4.10.2"
```
fixpoint: waitForProcess: does not exist ([smtlib-backends-process] while killing the solver process: No child processes)
```

In the circleci job:
```
using typed-process-0.2.8.0 and ghc-8.10.7
...
smtlib-backends-process          > [1 of 1] Compiling SMTLIB.Backends.Process
smtlib-backends-process          > 
smtlib-backends-process          > /tmp/stack-11b3cabb5dc5c281/smtlib-backends-process-0.2/src/SMTLIB/Backends/Process.hs:39:5: error:
smtlib-backends-process          >     Module ‘System.Process.Typed’ does not export ‘mkPipeStreamSpec’
smtlib-backends-process          >    |
smtlib-backends-process          > 39 |     mkPipeStreamSpec,
smtlib-backends-process          >    |     ^^^^^^^^^^^^^^^^
smtlib-backends-process          >
```
One solution is to specify a newer typed-process in the stack configuration, and to adjust the lower bound in the cabal file of `smtlib-backends-process`.

qaristote:
The failure in the CircleCI job probably comes from stack using smtlib-backends-process-0.2 and not 0.3 (I remember fixing the same error in the cabal files of smtlib-backends). So this should be fixed once we release smtlib-backends-0.3. In the meantime, I don't know whether it's possible to do something akin to the `cabal.project` and tell stack to get smtlib-backends from the GitHub repo.

qaristote:
https://github.com/ucsd-progsys/liquid-fixpoint/pull/641/commits/0cace8d6433523e98c8449a4e7d7cbb3c66ecb0b should provide a temporary fix to the CircleCI failure. I had to pin the version of typed-process as [stack was trying to use version 0.2.8.0 even though smtlib-backends' cabal file requests `typed-process >= 0.2.10`](https://app.circleci.com/pipelines/github/ucsd-progsys/liquid-fixpoint/680/workflows/5a25b8df-2793-4799-902f-daecba756fa1/jobs/1102?invite=true#step-106-44):
```
WARNING: Ignoring smtlib-backends-process's bounds on typed-process (>=0.2.10 && <0.3); using typed-process-0.2.8.0.
Reason: allow-newer enabled.
```

src/Language/Fixpoint/Smt/Interface.hs
```diff
@@ -303,75 +264,76 @@ makeContextWithSEnv cfg f env = do
...
 -- | Close file handles and wait for the solver process to terminate.
 cleanupContext :: Context -> IO ExitCode
-cleanupContext Ctx{..} = do
-  cancel ctxAsync
-  hCloseMe "ctxIn" ctxIn
-  hCloseMe "ctxOut"  ctxOut
+cleanupContext Ctx {..} = do
   maybe (return ()) (hCloseMe "ctxLog") ctxLog
-  waitForProcess ctxPid
+  ctxClose >> return ExitSuccess
```
qaristote: Should we also change the type of `cleanupContext` to `Context -> IO ()` ? Since it now only returns `ExitSuccess` and the return value doesn't seem to be used in the other modules.

facundominguez: LGTM

qaristote: https://github.com/ucsd-progsys/liquid-fixpoint/pull/641/commits/afd8c755fcb1f83f33de27c78c3a1d994579c157

stack.yaml
```diff
@@ -9,7 +9,6 @@ flags:
 packages:
 - '.'
 
-
 extra-deps:
 - hashable-1.3.5.0
 - rest-rewrite-0.3.0
```
facundominguez, line 14:
```suggestion
- rest-rewrite-0.3.0
- smtlib-backends-0.3
- smtlib-backends-z3-0.3
- smtlib-backends-process-0.3
```

facundominguez (approved):
Further measuring on benchmarks and tests reveals that asynchronous commands and `smtlib-backends` don't show up significant changes in performance.

Still, I think this integration is simplifying the interface to SMT solvers, and could give an advantage later on, when other parts of LF are optimized and become less dominant.

facundominguez:
My last measuring on tests of LH in `origin/develop` vs this PR:

perf_develop_vs_smtlibs_backends

PR #642, opened 2023-01-31
BitVector support

RobinWebbers:
This pull request extends liquid fixpoint to allow for arbirarily sized bitvectors as well as support for bitvector theories.

# Implementation

Currently I just extended the initial idea that was already there. This means that instead of type literals, I used the integer embedded within a type of form SizeX, (e.g. BitVec Size40 is a bitvector of 40 bits). I added most of the smt functions, such that we can do proofs using bvand, bvor, etc.

Apart from this, I removed some dead code regarding bitvectors (some type and parsing stuff). I've also added a test file that runs some of the added functions to show that it actually works: `tests/pos/bitvec.fq`

# Limitations (Help requested!)

So there are still some bitvector functions that are unavailable. These are the "indexed identifier" like functions in smtlib2 (e.g. `(_ sign_extend 4) bv`). However, I do really need support for these remaining functions from LF (and LH as well), so help on how to add this would be greatly appreciated!


ranjitjhala:
Wow, this is awesome thanks! May be best to meet so I can understand better how to make progress with this and #643 ?

RobinWebbers:
I've reached out to you via the slack PLSysSec channel as I thought it would be easier to communicate through there. Let me know if this works for you!

RobinWebbers:
So I added support for the indexed identifier functions, albeit in a somewhat weird way. I've basically tricked the evaluator into expanding function applications in an smtlib2 readable format. I've done this by implementing the functions `_` and `app`, which just apply their arguments to themselves. The elaborator will spit them out parenthesized correctly, where `app` is elaborated to the empty string when written to the smt solver.

To give an example for more clarity. The following expression in LF:
```
(app (_ sign_extend 4) bv)
```
is elaborated to the smt solver as
```
( (_ sign_extend 4) bv)
```
Which is exactly the format that the smt solver expects!

Notice how the `app` simply seems to disappear as it's mapped to the empty string `""`. The rest of this is just giving the correct types to `app`, `_` and the indexed identifier like functions such as `sign_extend` and `extract`.

The types that make the following typecheck are:
```
_ :: (a -> b -> c) -> a -> b -> c

app :: (a -> b) -> a -> b

sign_extend :: Int -> BitVec a -> BitVec a
```

I have yet to test how this exactly works out when used with Liquid Haskell btw. I expect it to just work, but who knows. I'll report back when I know!

ranjitjhala:
Oops sorry I thought we merged this - let me merge then rebase my stuff.

nmeum:
> I have yet to test how this exactly works out when used with Liquid Haskell btw. I expect it to just work, but who knows. I'll report back when I know!

Any updates on this? :)

RobinWebbers:
Hi. It's been a while, so I'm not 100% sure anymore, but I don't think it worked out of the box. If my memory serves well, the issue was that the embed syntax does not allow parametric embed statements.

Perhaps the following statement works:
```haskell
{-@ embed BitVec64 = (BitVec Size64) @-}
```

But I'm pretty sure the following wouldn't work:
```haskell
{-@ embed BitVec a = BitVec a @-}
```
Where `a` would then be embedded as a `SizeX` and applied to `BitVec`. Either way, I think it had to do with the embed statement itself as bitvector support seems to work in Liquid Fixpoint.

PR #644, opened 2023-02-10
`Language.Fixpoint.Smt.Interface`: replace parseWith with parseOnly

qaristote:
Following #641, the SMT solver's response is supplied in one go and the parser doesn't need the ability to read additional lines. This PR thus replace the call to `fmap A.eitherResult . A.parseWith :: IO Text -> Parser a -> Text -> IO (Either String a)` with a call to `parseOnly :: Parser a -> Text -> Either String a`.

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -31,6 +31,10 @@ jobs:
         with:
           version: ${{ matrix.z3 }}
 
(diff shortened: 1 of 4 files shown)


facundominguez (approved):

facundominguez:
Thanks @qaristote!

PR #655, opened 2023-09-28
Map test

ranjitjhala:
Adding ports of the tests for `Str` and `Map` to the horn format.


ranjitjhala:
@facundominguez -- any idea what's going wrong with the builds? I only added tests that seem to pass locally? looks like the `cabal ghc-9.2.3` build is using some newer version of the `Tasty`?

facundominguez:
The failing jobs are picking `tasty-1.5`, while stack is picking an older version. It may be an api change, indeed. Worth trying to put an upper bound to the tasty dependency.

Also, if you run `cabal update` in your machine, you might be able to reproduce it locally.

ranjitjhala:
Thanks that fixed it!

PR #661, opened 2023-11-19
Build liquid-fixpoint with ghc-9.4.7 in CI

facundominguez:


facundominguez:
`stan` is failing to run with ghc 9.4, so I'm keeping it at ghc 8.10.7 for now.
https://github.com/kowainik/stan/issues/541

PR #663, opened 2023-11-27
Stack project maintenance, remove allow-newer and more

philderbeast:
* Remove `allow-newer: true`
* Match flags between Cabal and Stack projects
* Be explicit with the version revisions
* Pick up `rest-rewrite` from stackage
* Bump the nightly resolver

Found when adding `liquid-fixpoint` as an [Updo example](https://github.com/up-do)

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,19 +1,18 @@
 
-resolver: nightly-2023-11-21
-allow-newer: true
+resolver: nightly-2023-11-27
 
 flags:
   liquid-fixpoint:
     devel: true 
+    link-z3-as-a-library: true
 
 packages:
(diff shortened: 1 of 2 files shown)


ranjitjhala:
LGTM! thanks @philderbeast ! -- @facundominguez does anything jump out to you? [the `link-z3-as-a-library: true` is still optional, yes? ]

facundominguez:
> is still optional, yes?

It is optional, but we want CI to test that it is actually optional. If both `cabal` and `stack` set the flag, we could customize CI to have some jobs disable the flag.

facundominguez (approved):

PR #666, opened 2023-12-04
Update Tuple definition to GHC 9.6 in the extensionality feature

facundominguez:


facundominguez:
I'll create a release of liquid-fixpoint to use with ghc-9.4.7.

PR #674, opened 2024-01-16
Use calculateCuts in elim to ignore cyclic constraints instead of crashing

K9-guardian:
Currently `Language.Fixpoint.Horn.Transformations.elim` crashes the program if the constraint contains cycles. As outlined in ICFP2017, the intended behavior should be to ignore the cyclic constraints and use predicate abstraction to solve them. This change uses `calculateCuts` to remove the set of refinement variables that cause the constraint to be cyclic before running the elimination algorithm.


ranjitjhala:
Thanks @K9-guardian !! I’m traveling atm but have started the CI … will take a closer look and merge when back in town tomorrow!

PR #678, opened 2024-02-09
Extend parser to allow boolean function arguments

ranjitjhala:
So that `f ?(x > y)` can be parsed as an `Expr` or `Pred`.

Ideally, we'd want to do this _without_ the nasty `?` but I can't seem to unravel the knots needed to do so :-(


ranjitjhala:
@facundominguez -- did you want to take a quick look at this, in case you have any suggestions?

facundominguez:
Hey, I'll be away from the keyboard for a couple of weeks. But yes, this
should be doable. If you need this merged, you could open a separate issue
to investigate removing the question mark.

PR #681, opened 2024-03-04
print SMT preamble to the logfile when constructing context + fix CircleCI

clayrat:
While investigating https://github.com/ucsd-progsys/liquidhaskell/issues/2272 we've discovered that since https://github.com/ucsd-progsys/liquid-fixpoint/pull/641 the SMT preamble is no longer printed to the `.smt2` logfile. This PR fixes that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,12 +4,12 @@ version: 2.0
 jobs:
   build:
     machine:
-      image: ubuntu-2004:202107-02
+      image: default
     steps:
       - run: sudo apt-get update && sudo DEBIAN_FRONTEND=noninteractive apt-get install -yq curl git ssh unzip wget libtinfo-dev gcc make
       - add_ssh_keys
-      - run: 
-          name: Install z3 
+      - run:
+          name: Install z3
           command: |
             wget https://github.com/Z3Prover/z3/releases/download/z3-4.8.7/z3-4.8.7-x64-ubuntu-16.04.zip
             unzip z3-4.8.7-x64-ubuntu-16.04.zip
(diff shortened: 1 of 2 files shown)


facundominguez:
~Thanks @clayrat, could we have a test that fails without this fix as well?~

facundominguez:
Ah, please, ignore my previous comment. This looks good as is.

facundominguez (approved):

ranjitjhala:
@facundominguez should we remove the `circleci` thing? Not sure why its failing but seems to be subsumed by the rest?

facundominguez:
I think we can drop the circleci job.

This job, however, was running liquid-fixpoint with the PLE interpreter on. If we are to continue testing the interpreter, we will need to run it elsewhere (maybe another github job).

clayrat:
I think the error just means that we have to update the image version in `.circleci/config.yml`

facundominguez:
> I think the error just means that we have to update the image version in .circleci/config.yml

Yes. I don't feel strong about fixing the job vs dropping it.

ranjitjhala:
@facundominguez do you recall why those `interpreter` tests are not just part of the regular `stack/cabal test` suite?
@clayrat can you try to change this line
https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/.circleci/config.yml#L7C14-L7C35

to say `image: default` ? Maybe that will work?

facundominguez:
> do you recall why those interpreter tests are not just part of the regular stack/cabal test suite?

I don't remember, but it is rerunning all tests with a different flag. If I had to guess, I'd say running a separate job was done to earn some parallelism. But the tests take only 30 seconds to run, so perhaps the parallelism is unjustified.

clayrat:
> @clayrat can you try to change this line [`develop`/.circleci/config.yml#L7C14-L7C35](https://github.com/ucsd-progsys/liquid-fixpoint/blob/develop/.circleci/config.yml?rgh-link-date=2024-03-04T20%3A26%3A46Z#L7C14-L7C35)
> to say `image: default` ? Maybe that will work?

Yep, works now

PR #683, opened 2024-03-11
Use SMTLIB style serialization/deserialization for Horn queries

ranjitjhala:
I want a human readable and machine serializable format for the horn queries, hence SMTLIB-style, to avoid all the knots of the FP parser.

Leaving the `Horn/Parse.hs` for backwards compatibility with `flux` guarded by the `--noHornSMT` flag; will delete it after moving `flux` over to this new format.

@facundominguez -- there was some odd glitch with the `--ple-with-undecided-guards` I had to remove the `-` as otherwise somehow it wasn't parsing? [ totally unrelated to this PR but it was breaking the CI! ]


src/Language/Fixpoint/Types/Config.hs
```diff
@@ -250,7 +251,7 @@ defConfig = Config {
   , rewriteAxioms            = False &= help "allow axiom instantiation via rewriting (PLE)"
   , pleWithUndecidedGuards   =
       False
-        &= name "ple-with-undecided-guards"
+        &= name "plewithundecidedguards"
```
facundominguez: Do you mean that keeping dashes here caused CI to fail? It would be helpful to see what the error is.

ranjitjhala: Yes keeping the dashes there made the intId test fail (locally) as CmdArgs kept saying “unknown flag”. More mysteriously, the test worked if I just wrote “—ple”…

ranjitjhala: I think it may be some strange MacOS thing ... have reverted the config flag name, will leave it be if it works in the CI!

ranjitjhala:
(I reverted the flag, will just deal with the local failure on my mac :-) )

PR #688, opened 2024-05-09
add internal Array theory, elaborate Set theory into it

### LINKED ISSUE #2272: Set measures generate erroneous constraints when applied to non-polymorphic datatypes
Typechecking the following program:

```haskell
{-@ LIQUID "--reflection" @-}

module SngBug where

import Data.Set

data Lst a = Emp | Cons a (Lst a)

{-@ measure lstHd @-}
lstHd :: Ord a => Lst a -> Set a
lstHd  Emp       = empty
lstHd (Cons x _) = singleton x

lcons ::  Lst l -> Lst (Lst l)
{-@ lcons :: p: Lst l -> {v:Lst (Lst l) | v = Cons p Emp } @-}
lcons p = Cons p Emp
```

crashes with the error:

`crash: SMTLIB2 respSat = Error "line 3 column 25217: unknown constant smt_set_sng ((SngBug.Lst Int)) declared: (declare-fun smt_set_sng (Int) (Array Int Bool)) "`

The error disappears if `lstHd` is declared `reflect` instead of `measure`. It triggers both with and without PLE.

It looks like the internal `Lst l` type doesn't get converted to `Int` which is what the encoding of sets as `Array Int Bool` expects. We currently think there are two possible solutions for this:

1. Only generate set operations for elements whose type can be encoded as `Int` (simpler solution, will turn a crash into a proper typechecking error)
2. Generate multiple set types for each potential set element type (more complex, but allows for expressing more programs)

clayrat:
Fixes https://github.com/ucsd-progsys/liquidhaskell/issues/2272 by elaborating Set operations into Array operations which are "internally" polymorphic in Z3. 

Currently, this only touches Sets but presumably Maps also need to undergo the same treatment.


clayrat:
Performance charts (on LH test suite):

bot
top
filtered

src/Language/Fixpoint/Smt/Theories.hs
```diff
@@ -63,24 +67,17 @@ import Language.Fixpoint.Utils.Builder
 -- | Theory Symbols ------------------------------------------------------------
 --------------------------------------------------------------------------------
 
+-- TODO drop all of Set and Map symbols when Map is handled through arrays
```
nikivazou, line 70: is this still TODO?

src/Language/Fixpoint/Smt/Theories.hs
```diff
@@ -507,16 +470,26 @@ theorySymbols ds = fromListSEnv $  -- SHIFTLAM uninterpSymbols
...
-  , interpSym setSng   sng  (FAbs 0 $ FFunc (FVar 0) (setSort $ FVar 0))
-  , interpSym setAdd   add   setAddSort
-  , interpSym setCup   cup   setBopSort
-  , interpSym setCap   cap   setBopSort
-  , interpSym setMem   mem   setMemSort
-  , interpSym setDif   dif   setBopSort
-  , interpSym setSub   sub   setCmpSort
-  , interpSym setCom   com   setCmpSort
+  [
+  -- TODO we'll probably need two versions of these - one for sets and one for maps
```
nikivazou, line 474: Still a TODO?

src/Language/Fixpoint/Types/Theories.hs
```diff
@@ -228,8 +228,10 @@ data SmtSort
   | SBool
   | SReal
   | SString
+  -- TODO remove these now that we use SArray directly
```
nikivazou, line 231: Can you now remove these?

nikivazou (commented):
Looks good to me! But does it come with a relevant LH pull request to ensure that it does not break any tests?

src/Language/Fixpoint/Smt/Theories.hs
clayrat: Yes, as currently the Map theory is still done via monomorphizing and it has operations for converting to Sets, we can't yet remove the SMT-interacting parts for Maps+Sets. Once Maps are also converted, we can drop this code and only leave the Array theory to be written to SMT. These TODOs serve as a reminder of this.

src/Language/Fixpoint/Types/Theories.hs
clayrat: Not yet, see above

src/Language/Fixpoint/Smt/Theories.hs
clayrat: Once we start converting Maps, there are likely to be some design choices whether to try making Array theory more polymorphic, or simply have a second set of Array operations specialized for Sets. This TODO is a heads-up for this.

facundominguez:
> by elaborating Set operations into Array operations which are "internally" polymorphic in Z3

Might require documentation: Does this PR narrow in some way the list of SMT solvers that LH/LF can work with?

clayrat:
> Might require documentation: Does this PR narrow in some way the list of SMT solvers that LH/LF can work with?

This uses the same SMTLIBv2-compliant ArraysEx theory encoding as before, we're just now inlining the definitions so that the LF elaborator can produce local signatures instead of hardcoding all Sets to be `Array Int Bool`. So in theory all the solvers that support SMTLIBv2 should still keep working.

However, the concern that by now we're pretty much vendor-locked into Z3 is valid; I've tried running the .smt files produced by LF even before this PR through CVC5 and failed; investigating this is our next step (after converting the Map theory to use Arrays as well).

clayrat:
> Looks good to me! But does it come with a relevant LH pull request to ensure that it does not break any tests?

The tests pass successfully after adapting LH to the `develop` branch: https://github.com/ucsd-progsys/liquidhaskell/pull/2282

nikivazou:
@ranjitjhala we can merge this tomorrow (just checking it does not break flux or anything else)

ranjitjhala:
Yes please merge, thanks @nikivazou  and @clayrat -- this is a really huge step!!!

ranjitjhala:
[I think its a big enough change to warrant a version number bump?]

nikivazou:
Alex wants to apply the same inlining for Maps (for uniformity) so we can change the version number after this pending edit.

PR #689, opened 2024-05-25
Remove allow newer

philderbeast:
* Bump `tested-with` GHC versions.
* Configure Stack workflow with resolver matching GHC versions.
* Bump dependencies so that `allow-newer` exceptions are no longer needed.


facundominguez:
Much appreciated!

PR #691, opened 2024-05-25
Bump to actions-checkout@v4

philderbeast:
Also remove submodules as we don't use those.


facundominguez:
Thanks @philderbeast!

PR #695, opened 2024-05-26
Bump to cabal-install-3.10.3.0

philderbeast:


facundominguez:
Thanks @philderbeast!

PR #696, opened 2024-06-17
refactor Set->Array elaboration

clayrat:
This PR compacts multiple local sort coercions introduced in #688 (currently for Set->Array, upcoming PR will also add Bag/Map->Array coercions) into a few recursive coercions on the arguments passed to instances of `Elaborate` and `Checkable` type classes. This reduces the possibility of some code paths not being covered by necessary coercions, e.g. when using PLE.

Additionally, this PR adds a more intuitive synonym `--ple` for `--rewriteaxioms` command line key that enables PLE.

Lastly, there's a change in the way errors are reported in [`Language.Fixpoint.Solver.crashResult`](https://github.com/clayrat/liquid-fixpoint/blob/ple-theories/src/Language/Fixpoint/Solver.hs#L189-L197) - previously this function would only display a generic `Sorry, unexpected panic in liquid-fixpoint!` error message if its error argument could not be found in the supplied `ErrorMap`; after the change this message will contain all available information, making debugging easier.


src/Language/Fixpoint/Solver.hs
```diff
@@ -190,8 +190,9 @@ crashResult :: (PPrint a) => ErrorMap a -> Error -> Result (Integer, a)
 crashResult m err' = Result res mempty mempty mempty
   where
     res           = Crash es msg
-    es            = catMaybes [ findError m e | e <- errs err' ]
-    msg | null es   = "Sorry, unexpected panic in liquid-fixpoint!"
+    es            = catMaybes [ findError m e | e <- ers ]
+    ers           = errs err'
+    msg | null ers = "Sorry, unexpected panic in liquid-fixpoint!"
```
nikivazou, line 195: can you explain this change?

nikivazou (commented):
Looks good to me!!! 

There is no performance check in the respective LH RP. Shall you make one?

src/Language/Fixpoint/Solver.hs
clayrat: updated the PR description with it

clayrat:
Added performance charts to https://github.com/ucsd-progsys/liquidhaskell/pull/2302. It seems we have significant slowdowns in ~5 tests, out of which `esop2013`, `bytestring`, `vector-algorithms` have large SMT encodings (10-100s Mb) and `ListSetDemo` is the only one actually using a Set encoding.

clayrat:
I've added an environment coercion to `isUnsat` to fix an SMT serialization error triggered in the new test added to LH; I haven't encountered this in any other test before.

PR #704, opened 2024-09-10
Unfold-unfolded definitions when the extentionality flag is active

AlecsFerra:
When trying to prove pretty trivial statements PLE gives up because is not unfolding enough the definitions or lambdas: see the following examples:

```haskell
{-@ LIQUID "--reflection"                @-}
{-@ LIQUID "--ple"                       @-}
{-@ LIQUID "--full"                      @-}
{-@ LIQUID "--extensionality"            @-}
{-@ LIQUID "--dump-opaque-reflections"   @-}
{-@ LIQUID "--ple-with-undecided-guards" @-}

module Example where

{-@ reflect f @-}
f :: Int -> Int
f x = 13

{-@ reflect g @-}
g :: Int -> Int -> Int
g x y = f y

{-@ reflect k @-}
k :: Int -> Int -> Int
k x y = 13


{-@ complexProof :: { f = (\x:Int -> 13) } @-}
complexProof :: ()
complexProof = ()

{-@ complexProof' :: { g = k } @-}
complexProof' :: ()
complexProof' = ()
````

### DIFF
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -861,15 +861,14 @@ evalApp γ ctx e0 es et
   , lambdaArg:remArgs <- es
   = do
       isFuelOk <- checkFuel argName
-      ext <- isExtensionalityOn
-      if isFuelOk && ext 
+      isExtensionalityOn <- gets extensionalityFlag
+      if isFuelOk && isExtensionalityOn
         then do
           useFuel argName
           let argSubst = mkSubst [(argName, lambdaArg)]
           let body' = subst argSubst body
           (body'', fe) <- evalIte γ ctx et body'
           let simpBody = simplify γ ctx (eApps body'' remArgs)
-          -- This is still the same thing I'm doing in the application of functions
           modify $ \st ->
             st { evNewEqualities = S.insert (eApps e0 es, simpBody) (evNewEqualities st) }
           return (Just $ eApps body'' remArgs, fe)
@@ -1228,9 +1227,6 @@ useFuelCount f fc = fc { fcMap = M.insert f (k + 1) m }
   where
     k             = M.lookupDefault 0 f m
     m             = fcMap fc
-
-isExtensionalityOn :: EvalST Bool
-isExtensionalityOn = gets extensionalityFlag
 
 -- | Returns False if there is a fuel count in the evaluation environment and
 -- the fuel count exceeds the maximum. Returns True otherwise.
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -146,7 +146,6 @@ instEnv cfg info cs restSolver ctx = do
               , evNewEqualities = mempty
               , evSMTCache = mempty
               , evFuel = defFuelCount cfg
-              , unfoldedDefinitions = mempty
               , extensionalityFlag = extensionality cfg
               , explored = Just et
               , restSolver = restSolver
@@ -460,7 +459,6 @@ data EvalEnv = EvalEnv
                                     -- an expression
   , evSMTCache :: M.HashMap Expr Bool -- ^ Whether an expression is valid or its negation
   , evFuel     :: FuelCount
-  , unfoldedDefinitions :: S.HashSet Symbol -- ^ Functions that we have already unfolded
   , extensionalityFlag :: Bool -- ^ True if the extensionality flag is turned on
 
   -- REST parameters
@@ -788,10 +786,9 @@ evalApp γ ctx e0 es et
   , Just eq <- Map.lookup f (knAms γ)
   , length (eqArgs eq) <= length es
   = do
-       alreadyUnfolded <- isAlreadyUnfolded f
        env  <- gets (seSort . evEnv)
        okFuel <- checkFuel f
-       if okFuel && et /= FuncNormal && not alreadyUnfolded
+       if okFuel && et /= FuncNormal
          then do
                 let (es1,es2) = splitAt (length (eqArgs eq)) es
                     newE = substEq env eq es1
@@ -801,12 +798,6 @@ evalApp γ ctx e0 es et
                     undecidedGuards = case e' of
                       EIte{} -> True
                       _ -> False
-
-                ext <- isExtensionalityOn
-                when ext $ do
-                  setUnfolded f
-                  _ <- eval γ ctx et e3'
-                  unsetUnfolded f
 
                 if undecidedGuards
                   then do
@@ -825,7 +816,7 @@ evalApp γ ctx e0 es et
                         { evNewEqualities = S.insert (eApps e0 es, e3') (evNewEqualities st)
                         , evPendingUnfoldings = M.delete (eApps e0 es) (evPendingUnfoldings st)
                         }
-                    return (Just e2', fe)
+                    return (Just $ eApps e2' es2, fe)
          else return (Nothing, noExpand)
   where
     -- At the time of writing, any function application wrapping an
@@ -1239,24 +1230,10 @@ useFuelCount f fc = fc { fcMap = M.insert f (k + 1) m }
     k             = M.lookupDefault 0 f m
     m             = fcMap fc
 
-setUnfolded :: Symbol -> EvalST ()
-setUnfolded sym = modify 
-  (\st -> st { unfoldedDefinitions = S.insert sym (unfoldedDefinitions st)})
-
-isAlreadyUnfolded :: Symbol -> EvalST Bool
-isAlreadyUnfolded sym = do
-    st <- get
-    return $ S.member sym $ unfoldedDefinitions st
-
 isExtensionalityOn :: EvalST Bool
 isExtensionalityOn = do
     st <- get
     return $ extensionalityFlag st
-
-
-unsetUnfolded :: Symbol -> EvalST ()
-unsetUnfolded sym = modify 
-  (\st -> st { unfoldedDefinitions = S.delete sym (unfoldedDefinitions st)})
 
 -- | Returns False if there is a fuel count in the evaluation environment and
 -- the fuel count exceeds the maximum. Returns True otherwise.
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -1230,9 +1230,7 @@ useFuelCount f fc = fc { fcMap = M.insert f (k + 1) m }
     m             = fcMap fc
 
 isExtensionalityOn :: EvalST Bool
-isExtensionalityOn = do
-    st <- get
-    return $ extensionalityFlag st
+isExtensionalityOn = gets extensionalityFlag
 
 -- | Returns False if there is a fuel count in the evaluation environment and
 -- the fuel count exceeds the maximum. Returns True otherwise.
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -1231,9 +1231,7 @@ useFuelCount f fc = fc { fcMap = M.insert f (k + 1) m }
     m             = fcMap fc
 
 isExtensionalityOn :: EvalST Bool
-isExtensionalityOn = do
-    st <- get
-    return $ extensionalityFlag st
+isExtensionalityOn = gets extensionalityFlag
 
 -- | Returns False if there is a fuel count in the evaluation environment and
 -- the fuel count exceeds the maximum. Returns True otherwise.
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -867,13 +867,12 @@ evalApp γ ctx e0 es et
           useFuel argName
           let argSubst = mkSubst [(argName, lambdaArg)]
           let body' = subst argSubst body
-          (body'', _) <- evalIte γ ctx et body'
+          (body'', fe) <- evalIte γ ctx et body'
           let simpBody = simplify γ ctx (eApps body'' remArgs)
           -- This is still the same thing I'm doing in the application of functions
-          (body''', fe) <- eval γ ctx et simpBody
           modify $ \st ->
-            st { evNewEqualities = S.insert (eApps e0 es, body''') (evNewEqualities st) }
-          return (Just body'', fe)
+            st { evNewEqualities = S.insert (eApps e0 es, simpBody) (evNewEqualities st) }
+          return (Just $ eApps body'' remArgs, fe)
         else do
           return (Nothing, noExpand)
 
@@ -1231,7 +1230,9 @@ useFuelCount f fc = fc { fcMap = M.insert f (k + 1) m }
     m             = fcMap fc
 
 isExtensionalityOn :: EvalST Bool
-isExtensionalityOn = gets extensionalityFlag
+isExtensionalityOn = do
+    st <- get
+    return $ extensionalityFlag st
 
 -- | Returns False if there is a fuel count in the evaluation environment and
 -- the fuel count exceeds the maximum. Returns True otherwise.
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -146,6 +146,8 @@ instEnv cfg info cs restSolver ctx = do
               , evNewEqualities = mempty
               , evSMTCache = mempty
               , evFuel = defFuelCount cfg
+              , unfoldedDefinitions = mempty
+              , extensionalityFlag = extensionality cfg
               , explored = Just et
               , restSolver = restSolver
               , restOCA = restOrd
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by AlecsFerra** [comment]
I have ran all the tests in the liquid-haskell repo and they are passing

**Comment by facundominguez** [comment]
:wave: Hello @AlecsFerra. The following works for me.
```Haskell
{-@ LIQUID "--ple"                       @-}

module Example where

{-@ reflect f @-}
f :: Int -> Int
f x = 13

{-@ reflect g @-}
g :: Int -> Int -> Int
g x y = f y

{-@ reflect k @-}
k :: Int -> Int -> Int
k x y = 13


{-@ complexProof' :: x:Int -> y:Int -> { g x y = k x y } @-}
complexProof' :: Int -> Int -> ()
complexProof' _ _ = ()
```
Enabling `--extensionality` seems to break it, though.

I don't know how to deal with the lambda in the first lemma. But my intuition is that functions like `f` need to be applied to actual arguments before they can be unfolded.

AlecsFerra:
The thing is that the extensionality flag is adding some "fake" variables that prevent the unfoldings

AlecsFerra:
So it should pass without adding the extra arguments to the proof

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -458,6 +460,8 @@ data EvalEnv = EvalEnv
                                     -- an expression
   , evSMTCache :: M.HashMap Expr Bool -- ^ Whether an expression is valid or its negation
   , evFuel     :: FuelCount
+  , unfoldedDefinitions :: S.HashSet Symbol -- ^ Functions that we have already unfolded
```
facundominguez: Thinking about documentation:

`unfoldedDefinitions` seems to be used to inhibit unfolding of some saturated function application, for instance `length xs`. However, it would also inhibit unfolding of `length ys`. Is this necessary?

Also, the scope in which a function is inhibited from unfolding needs to be made explicit. It seems to be inhibited only when scanning for more applications to unfold, in the result of the first unfolding (`eval γ ctx et e3'` below). Why?

facundominguez (commented):
Thanks for the PR.

I need some help understanding the problem, I think. My comment below contains my first questions.

Also, before merging, looks like some new tests are going to be needed here and in liquidhaskell.

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Technically we could keep in the unfolded set of also arguments of the call `(Symbol, [Expr])` but I'm worried that it could be tricked to end in an infinite loop somehow since the termination checking happens after this code is ran

I don't understand your second comment sorry

AlecsFerra:
The problem is the following:

morally
```haskell
{-@ complexProof' :: x:Int -> y:Int -> { g x y = k x y } @-}
complexProof' :: Int -> Int -> ()
complexProof' _ _ = ()
```
and
```haskell
{-@ complexProof' :: { g = k } @-}
complexProof' :: ()
complexProof' = ()
```

Have two different statements. we would like to LH to prove by itself the second statement as it's a common class of proofs that we have to do when performing higher order reasoning in LH.

As part of a project I'm proving the correctness of compiling terms of the lambda calculus to the SKI combinators:

See: https://gist.github.com/AlecsFerra/9d48ae08f63d8192b2267d0e378f8c09
You can see the file Try3.hs using the old PLE and Whole.hs using the new PLE
the proofs are much simpler (Try3.hs has the proof of only some of the statements, instead Whole.hs has a proof of all the statements)

AlecsFerra:
For the proofs I wanted to wait for some comments about the PR (both on the implementation  and on the idea) hence why is marked as a draft, I've already discussed about this with @nikivazou

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: > but I'm worried that it could be tricked to end in an infinite loop somehow

How would PLE enter an infinite loop if `unfoldedDefinitions` isn't introduced at all?

> I don't understand your second comment sorry

This ties to the motivation of `unfoldedDefinitions` which isn't clear to me yet. Symbols aren't kept in `unfoldedDefinitions` forever, they are removed with `unsetUnfolded`. And I'm asking how did you choose when to call `unsetUnfolded`.

> For the proofs I wanted to wait for some comments about the PR (both on the implementation and on the idea) hence why is marked as a draft,

I hadn't noticed this PR was marked as a draft. Collecting feedback sounds good.

AlecsFerra:
Also forgot to mention this code works even if we drop the check for the extensionality flag, I've already tested running the tests in the main liquidhaskell repo, is just making the typechecker slower

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Take as an example

```haskell
{-@ LIQUID "--reflection"                @-}
{-@ LIQUID "--ple"                       @-}
{-@ LIQUID "--full"                      @-}
{-@ LIQUID "--extensionality"            @-}
{-@ LIQUID "--ple-with-undecided-guards" @-}

module Example where

{-@ reflect f @-}
f :: Int -> Int
f n = if n > 0 then f n + n else 0

{-@ reflect g @-}
g :: Int -> Int -> Int
g x y = f 3


{-@ complexProof :: { g = (\x:Int -> (\y:Int -> 3)) } @-}
complexProof :: ()
complexProof = ()

```

Even tough LH is able to infer that f is non terminating, LH will try to look for some other errors so when trying to prove `complexProof` will start with the unfolding of g so it will try to unfold f 3 so it will try to unfold then then branch of the body of f and so on till we finish the space for the call stack.

You can try it by your self by commenting away L807.

AlecsFerra: Cool thing :) we have infinite loops only in non terminating programs we dont have infinite loops in programs that are safe

```haskell
{-@ LIQUID "--reflection"                @-}
{-@ LIQUID "--ple"                       @-}
{-@ LIQUID "--full"                      @-}
{-@ LIQUID "--extensionality"            @-}
{-@ LIQUID "--ple-with-undecided-guards" @-}

module Example where

{-@ reflect f @-}
f :: Int -> Int
f n = if n > 0 then f (n - 1) + n else 0

{-@ reflect g @-}
g :: Int -> Int -> Int
g x y = f 3


{-@ complexProof :: { g = (\x:Int -> (\y:Int -> 6)) } @-}
complexProof :: ()
complexProof = ()
```

For example this passes even without keeping track of the unfolded definitions

AlecsFerra: Regarding `unsetUnfolded`, we have to remove them from the set since we sometimes want to use the unfolding of the function in 2 different proof maybe in even different scopes

facundominguez: Thanks. I had to add `{-@ lazy f @-}` to have LH accept that example.

I can see that it enters an infinite loop, and I can see why when commenting line 807. However, it passes verification if I also comment line 808 which is new, and is participating in the loop. Can the example be modified so line 808 is actually needed for verification to succeed?

facundominguez: > Can the example be modified so line 808 is actually needed for verification to succeed?

Nevermind, the example in the description of the PR seems to be one such example. :thinking:

AlecsFerra: Line 808 is what is performing the "deep" unfolding.

Lets take
```haskell
{-@ complexProof' :: { g = k } @-}
complexProof' :: ()
complexProof' = ()
```

The extensionality pass will change the proof in
```haskell
{-@ complexProof' :: { g ext$0 ext$1 = k ext$0 ext$1 } @-}
complexProof' :: ()
complexProof' = ()
```

Now what happens with the original PLE is that we have one layer of unfolding
```haskell
{-@ complexProof' :: { f ext$1 = 13 } @-}
complexProof' :: ()
complexProof' = ()
```
For`k` is enough but for `g` is not we need to unfold also `f ext$1` but the original PLE performs one unfolding ad it stops

Instead the new PLE at line 808 calls eval again and will unfold `f ext$1` to 13 making the proof automatic by LH

facundominguez: > but the original PLE performs one unfolding ad it stops

I think that is a bug in PLE, it shouldn't stop. The following patch fixes an omission in the existing code, by also proving redundant the recursive call to `eval γ ctx et e3'`.

```
diff --git a/src/Language/Fixpoint/Solver/PLE.hs b/src/Language/Fixpoint/Solver/PLE.hs
index cc4644d1..1740106b 100644
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -805,7 +805,7 @@ evalApp γ ctx e0 es et
                 ext <- isExtensionalityOn
                 when ext $ do
                   setUnfolded f
-                  _ <- eval γ ctx et e3'
+                  -- _ <- eval γ ctx et e3'
                   unsetUnfolded f
 
                 if undecidedGuards
@@ -825,7 +825,7 @@ evalApp γ ctx e0 es et
                         { evNewEqualities = S.insert (eApps e0 es, e3') (evNewEqualities st)
                         , evPendingUnfoldings = M.delete (eApps e0 es) (evPendingUnfoldings st)
                         }
-                    return (Just e2', fe)
+                    return (Just $ eApps e2' es2, fe)
          else return (Nothing, noExpand)
   where
     -- At the time of writing, any function application wrapping an
```

PLE cannot continue if `evalApp` doesn't return the correct unfolding of the expression `eApp e0 es`. 

But this is not the end of the story for me. When tracing the above example, I found that the reflection of `g` is amiss. In the .fq file, it looks correct:
```
define Example.g (lq1 : int,  lq2 : int) : int = {((Example.f lq2))}
```
but in the `eq` variable in `evalApp` it is shows as:
```
Equ
  { eqName = "Example.g"
  , eqArgs = [("lq1##Example.g##0",FInt)]
  , eqBody = EVar "Example.f"
  , eqSort = FFunc FInt FInt, eqRec = False
  }
```
It is missing the "lq2" argument, and the eqBody should be a function application, not a variable. Has `liquid-fixpoint` eta reduced the reflection??

AlecsFerra: `return (Just $ eApps e2' es2, fe)` makes LH perform the right amount of unfolding so that if able to verify `g = k`? How does it handles the infinite loop? I have turned out my PC so I cant test it right now but it marks `Whole.hs` as safe?

facundominguez: > Has liquid-fixpoint eta reduced the reflection??

Ah, clarified the bit on eta elimination. Yes, LF is doing eta elimination. The .fq file has a flag `etaElim` enabled when LH calls LF.

facundominguez: > How does it handles the infinite loop?

PLE keeps a list of unfoldings that have been discovered (`icEquals`). I guess the infinite loop is avoided because it is crafted by always unfolding the same expression over and over. Once the unfolding enters `icEquals`, PLE doesn't continue.

I tried all the examples in this PR, and they all seem to pass.

AlecsFerra: Can you try Whole.hs in this gist? https://gist.github.com/AlecsFerra/9d48ae08f63d8192b2267d0e378f8c09 it is rejected by the old PLE but accepted by mine

Note that we still need the code that I added for the evaluation of lambdas

facundominguez: I can still make PLE diverge with
```
{-@ reflect f @-}
{-@ lazy f @-}
f :: Int -> Int
f n = if n > 0 then f (n+1) + n else 0

{-@ reflect g @-}
g :: Int -> Int -> Int
g x y = f 3

{-@ complexProof :: { g = (\x:Int -> (\y:Int -> 3)) } @-}
complexProof :: ()
complexProof = ()
```
but same as before, one still needs to use `lazy` which is an unsafe thing to do when using the function in proofs.

AlecsFerra: If its the case then I guess we can perform the same modification in the case I added at line 868, removing the recursive call call and modifying the return

facundominguez: > we can perform the same modification in the case I added at line 868

The return in that case looks correct to me.

AlecsFerra: Yeah but I'm calling eval inside that case maybe it isn't needed if PLE decides by looking at what is returned it it wants to unfold again

AlecsFerra: Btw regarding making PLE diverge i think its our job to make LH terminate anyway, also before I added the infinite loop check, some test in the liquidhaskell repo was entering in the same infinite loop

nikivazou:
Great! Let's not merge yet then, and keep it as draft until we come up with more tests. 

`hlint` also complains :) 

The tests on LH should not be slower if the extra unfoldings happen only when the extensionality flag is on, correct?

AlecsFerra:
After the discussion with @facundominguez yesterday the fact that PLE wasn't unfolding recursively was a bug in PLE, so I decided to deleted my code that was dealing with function application and implement the fix that was suggested in the comments above.
That solutions leads PLE in an infinite loop in some cases (I personally believe that we should stop PLE like I was doing in my first commit).
Then I dont know if it still makes sense to hide this feature behind the extensionality flag, I'll run the benchmarks this afternoon.

I left the code that handles lambdas untouched and behind the extensionality flag.

AlecsFerra:
Let me know if i should squash the commits

AlecsFerra:
Regarding tests I think that we can reuse all the snippets in the discussion + the whole SKI compilation file for the liquid haskell main repo.

For the liquid fixpoint repo first I have to study the file format

AlecsFerra:
Benchmarks:

filtered
top
bot

AlecsFerra:
I modified the case for the lambda application to make it behave the same as normal function application, I don't know if we have to handle the undecided guards thing

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -854,6 +856,26 @@ evalApp γ ctx e0 es _et
        modify $ \st ->
          st { evNewEqualities = foldr S.insert (evNewEqualities st) eqs' }
        return (Nothing, noExpand)
+evalApp γ ctx e0 es et
+  | ELam (argName, _) body <- dropECst e0
+  , lambdaArg:remArgs <- es
+  = do
+      isFuelOk <- checkFuel argName
+      ext <- isExtensionalityOn
+      if isFuelOk && ext
```
```suggestion
      if isFuelOk && ext
```

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -854,6 +856,26 @@ evalApp γ ctx e0 es _et
...
+      isFuelOk <- checkFuel argName
+      ext <- isExtensionalityOn
+      if isFuelOk && ext 
+        then do
+          useFuel argName
+          let argSubst = mkSubst [(argName, lambdaArg)]
+          let body' = subst argSubst body
+          (body'', fe) <- evalIte γ ctx et body'
+          let simpBody = simplify γ ctx (eApps body'' remArgs)
+          -- This is still the same thing I'm doing in the application of functions
```
```suggestion
```

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Maybe inlining `isExtensionalityOn` is still readable enough, and it would free the name to be used for the resulting boolean.
```suggestion
      isExtensionalityOn <- gets extensionalityFlag
      if isFuelOk && isExtensionalityOn
```

facundominguez (commented):
The code is looking good to me.

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Happy to discuss that further, but I don't think it affects the contribution in this PR. I'd propose to open a separate issue or PR for that sake.

facundominguez:
> I don't know if we have to handle the undecided guards thing

Edited this answer after posting, see next comment.

facundominguez:
hm, `evPendingUnfoldings` does rather delay adding new equalities that may not be insightful. In `length ys`, most of the time it is not helpful to add an equality
```Haskell
length ys = if null ys then 0 else length ys + 1
```
unless the `null ys` condition can be decided.

For lambda expressions
```Haskell
(\xs -> if null xs then e0 else e1) ys
```
I think the question is: does it help most of the time to add an equality
```Haskell
(\xs -> if null xs then e0 else e1) ys = if null ys then e0 else e1
```
when the condition `null ys` cannot be decided? If it does help, then leave the code as is. If it doesn't, then imitate the non-lambda case.

AlecsFerra:
No It helps as that equality is encoding "eta-equivalence" that before my PR it wasn't provable in LH

AlecsFerra:
How do I get the .fq files? I can't find them in the `.liquid` directory.
I want them so that I can add the examples as tests in liquid-fixpoint

facundominguez:
> How do I get the .fq files?

`{-@ LIQUID "--save" @-}`

AlecsFerra:
Let me know if those tests are enough

nikivazou (approved):

PR #705, opened 2024-09-13
Extensional reasoning on Values

AlecsFerra:
The extensionality flag fails when a function used as argument of a data constructor, take this as an example:

```
fixpoint "--rewrite"
fixpoint "--allowho"

constant f : (func(0 , [int; int; int]))
define f (x : int, y: int) : int = {(x + y)}

constant g : (func(0 , [int; int; int]))
define g (a : int, b: int) : int = {(b + a)}


data Ty 0 = [
    | Cons {mkCons : func(0 , [int; int; int])}
]

constant Cons : (func(0 , [func(0 , [int; int; int]); Ty]))

expand [1 : True; 2 : True]

constraint:
  env []
  lhs {VV1 : Tuple | true }
  rhs {VV2 : Tuple | (Cons f = Cons g) }
  id 2 tag []
```

it cannot prove that f and g are the same function since it cannot add the `ext$...` variable on the functions (and then cause their unfolding by PLE) since they would make the program ill-typed, the solution is eta-expanding the definitions of `f` and `g` to `(\x:Int -> \y:Int -> f x y)` `(\x:Int -> \y:Int -> g x y)` to let PLE then unfold the definitions in the body of the lambda and prove the equality.

DO NOT MERGE: This feature is exposing another LF bug in some tests in the liquidhaskell repo related to lambdas in the refinements.


AlecsFerra:
The issue on the liquid haskell main repo is related to programs like this:
```haskell
{-@ LIQUID "--reflection" @-}
{-@ LIQUID "--ple"        @-}

module Example where

{-@ reflect idu @-}
{-@ idu :: Int -> Int @-}
idu :: Int -> Int
idu x = x

{-@ proof :: {(\x:Int -> x) = idu } @-}
proof = ()
```

As it throws `crash: SMTLIB2 respSat = Error "line 3 column 18650: unknown constant lam_arg$35$$35$1"` whenever a lambda is used this way in the refinements

This error is already present is not caused by this patch

nikivazou:
Yes, right now there is no real support for lambdas in LH :)

AlecsFerra:
Uhm ok then I'm stuck

AlecsFerra:
Right now It's verifying the example but it's also making some tests in the LH repo fail because it's unfolding some types and inserting lambdas

AlecsFerra:
I think I fixed the bug regarding lambdas, I'll run the tests on the main repo to double check

AlecsFerra:
The tests on the liquidhaskell are passing!

But I still have to fix a minor bug for my tests

src/Language/Fixpoint/Solver/Common.hs
```diff
@@ -16,7 +16,7 @@ askSMT :: Config -> Context -> [(Symbol, Sort)] -> Expr -> IO Bool
 askSMT cfg ctx xs e
 --   | isContraPred e  = return False
   | isTautoPred  e     = return True
-  | null (kvarsExpr e) = checkValidWithContext ctx [] PTrue e'
+  | null (kvarsExpr e) = checkValidWithContext ctx xs PTrue e'
```
facundominguez, line 19: I don't understand this code too well, but it is around since 2018 or earlier. Can we have a small test that fails without this change?

One problem here is that there is no documentation to what `xs` is meant to be. If we are to infer it from the existing code, it is the environment used for elaboration of the expression.

This code is changing the meaning of `xs` to also declare the symbols to the SMT solver in addition to evaluating the expression. Is this change in meaning necessary?

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -564,8 +574,8 @@ feSeq xs = (map fst xs, feAny (map snd xs))
 -- Also adds to the monad state all the unfolding equalities that have been
 -- discovered as necessary.
 --
-eval :: Knowledge -> ICtx -> EvalType -> Expr -> EvalST (Expr, FinalExpand)
-eval γ ctx et = go
+eval :: Knowledge -> ICtx -> EvalType -> Bool -> Expr -> EvalST (Expr, FinalExpand)
+eval γ ctx et isLHSApp = go
```
facundominguez: Please, could we have the meaning of this new parameter documented?

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -603,9 +615,13 @@ eval γ ctx et = go
...
     go (POr es)         = efAll POr (go `traverse` es)
     go e | EVar _ <- dropECst e = do
-      (me', fe) <- evalApp γ ctx e [] et
-      return (Mb.fromMaybe e me', fe)
-    go (ECst e t)       = do (e', fe) <- eval γ ctx et e
+                                    if not isLHSApp then do
+                                    (me', fe) <- evalApp γ ctx e [] et
+                                    return (Mb.fromMaybe e me', fe)
+                                    else do
+                                    return (e, noExpand)
```
facundominguez: Please, could we have a comment explaining the need for the different cases?

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -625,15 +641,15 @@ evalELam :: Knowledge -> ICtx -> EvalType -> (Symbol, Sort) -> Expr -> EvalST (E
...
+    (e', fe) <- eval (γ { knLams = (x, s) : knLams γ }) ctx et False e
     let e2' = simplify γ ctx e'
         elam = ELam (x, s) e
     -- Discard the old equalities which miss the lambda binding
     modify $ \st -> st
       { evPendingUnfoldings = oldPendingUnfoldings
       , evNewEqualities = S.insert (elam, ELam (x, s) e2') oldEqs
       }
-    return (elam, fe)
+    return (ELam (x, s) e2', fe)
```
facundominguez: Is this change necessary?

I have vague recollections that we don't simplify the return values to avoid missing potential rewrites. This change is setting a different criterion for the lambda case.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -879,11 +939,15 @@ evalApp γ ctx e0 es et
 evalApp _ _ _e _es _
   = return (Nothing, noExpand)
 
+
+wrap :: Expr -> [(Symbol, Sort)] -> Expr
```
facundominguez: It is not a very telling name. `wrap` is called `mkLams` in the GHC API.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -460,6 +464,12 @@ data EvalEnv = EvalEnv
   , evSMTCache :: M.HashMap Expr Bool -- ^ Whether an expression is valid or its negation
   , evFuel     :: FuelCount
   , extensionalityFlag :: Bool -- ^ True if the extensionality flag is turned on
+  
+  -- Eta expansion feature
+  , freshEtaNames :: Int -- ^ Keeps track of how many names we generated to perform eta 
+                         --   expansion, we use this to generate always fresh names
+  , hoFlag        :: Bool -- ^ True if the allowho flag is turned on, needed for the eta
+                          -- expansion reasoning as its going to generate ho constraints
```
facundominguez: This comment would need to point to a more precise explanation of what such reasoning is. It could be done with a longer note at the top-level.

From what I can see, it is just enabling beta reduction of lambdas and eta expansion of functions which appear in non-saturated applications.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -830,6 +847,47 @@ evalApp γ ctx e0 es et
       , f == "Language.Haskell.Liquid.ProofCombinators.pleUnfold"
       = arg
       | otherwise = e
+evalApp γ _ctx e0 es _et
+  | EVar f <- dropECst e0
+  , Just eq <- Map.lookup f (knAms γ)
+  = do
+    isHOAllowed <- gets hoFlag
+    if isHOAllowed then do
+    let expectedArgs = eqArgs eq
```
facundominguez: Indentation is amiss. It should be nested at least two further spaces to the right.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -867,7 +866,19 @@ evalApp γ _ctx e0 es _et
         -- This case is pretty sus as we shouldnt generate funcitons that do not
         -- have a type annotation
         Nothing -> do
-            traceShowM ("I'm failing on" :: String, f)
+          let etaArgsType = fmap snd $ drop nProvidedArgs $ eqArgs eq
+          -- Hack to recover general mono signature
+          if not $ isAnyTyVar etaArgsType then do
+          let etaVars = zipWith (\name ty -> ECst (EVar name) ty) etaNames etaArgsType
+          let fullBody = eApps e0 (es ++ etaVars)
+          let etaExpandedTerm = wrap fullBody (zip etaNames etaArgsType)
```
facundominguez: Looks like defining an auxiliary function for eta expansion could eliminate some duplication.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -867,7 +866,19 @@ evalApp γ _ctx e0 es _et
         -- This case is pretty sus as we shouldnt generate funcitons that do not
         -- have a type annotation
         Nothing -> do
-            traceShowM ("I'm failing on" :: String, f)
+          let etaArgsType = fmap snd $ drop nProvidedArgs $ eqArgs eq
+          -- Hack to recover general mono signature
+          if not $ isAnyTyVar etaArgsType then do
+          let etaVars = zipWith (\name ty -> ECst (EVar name) ty) etaNames etaArgsType
```
facundominguez: Indentation is amiss.

facundominguez (commented):
Hello @AlecsFerra. The changes make some sense to me. I contributed some questions and comments.

src/Language/Fixpoint/Solver/Common.hs
AlecsFerra: I think it was impossible to trigger a crash before my change, xs is a list containing all the variables in scope, so they are actually needed when checking that e' is valid as e' it can reference some variables in the outer scope

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: No it shouldn't be needed, I'll revert the change

AlecsFerra: But I need to test it before

AlecsFerra: No It's actually needed, since now we need to keep track that the body was rewritten

facundominguez: In that case, I suggest to use the unsimplified `e'` instead of `e2'`.

src/Language/Fixpoint/Solver/Common.hs
facundominguez: > so they are actually needed when checking that e' is valid as e'

These might have been already given to the smt solver before calling `askSMT`, or they might not be needed. Otherwise, tests would be failing because of undeclared identifiers.

I think there is still not enough of a motivation for this change.

nikivazou: If I remember correctly (and from the calls of this function) this `xs` indeed is meant to give to the SMT the lambda arguments. Since up to now LH does not really have support for lambdas, it makes sense that this was overlooked. 

If @AlecsFerra needs this edit to make this code work and it does not make any of the existing tests failing, I suggest we leave it.

src/Language/Fixpoint/Solver/PLE.hs
nikivazou: I agree that we can miss potential rewrites at each selection. But, again, since this is the first real work (after the initial old attempt) to make reasoning with lambdas work, and there is an example that actually needs it, I suggest we accept.

src/Language/Fixpoint/Solver/Common.hs
AlecsFerra: If you try to run this https://gist.github.com/AlecsFerra/c31162353b21e034dc3797c5c227c049 without passing xs its going to crash with error `  crash: SMTLIB2 respSat = Error "line 28 column 3941: unknown constant eta$35$$35$0"`

AlecsFerra:
Can confirm that it's passing all the test on the main LH repo, now I'm running the benchmarks

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: I replaced it with e'

AlecsFerra:
Benchmarks:
filtered
top
bot

On average we are not extremely slower than then original PLE

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -846,51 +859,53 @@ evalApp γ ctx e0 es et
...
+      -- the equation
+      let etaArgsMono = drop nProvidedArgs <$> extractMonoSign e0
+      case etaArgsMono of
+          -- This case is pretty sus as we shouldnt generate funcitons that do not
+          -- have a type annotation
+          Nothing -> do
+            let etaArgsType = fmap snd $ drop nProvidedArgs $ eqArgs eq
+            -- Hack to recover general mono signature
+            if not $ isAnyTyVar etaArgsType then do
+              performEtaExpansion e0 es etaNames etaArgsType
```
facundominguez: How about calling `makeFreshEtaNames` inside `performEtaExpansion`? The result of `makeFreshEtaNames` seems to be used only in this function.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -830,6 +857,81 @@ evalApp γ ctx e0 es et
...
+
+      -- Fresh names for the eta expansion
+      etaNames <- makeFreshEtaNames nArgsMissing
+      -- For some reason the application is expected to be already 
+      -- monomorphized, so we can't extract the sorts directly form
+      -- the equation
+      let etaArgsMono = drop nProvidedArgs <$> extractMonoSign e0
+      case etaArgsMono of
+          -- This case is pretty sus as we shouldnt generate funcitons that do not
+          -- have a type annotation
```
facundominguez: Typos: funcitos and sus (or at least I don't know what sus is)

The comment says that functions should have type annotations, but I guess there are cases were they may not. Otherwise, the implementation here would be an error call or similar.

Can we have an example to show how to rehearse this case?

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -830,6 +857,81 @@ evalApp γ ctx e0 es et
...
+          Nothing -> do
+            -- Hack to recover general mono signature, by looking directly
+            -- at the types in the equation
+            let etaArgsType = fmap snd $ drop nProvidedArgs $ eqArgs eq
+            if not $ any hasTyVar etaArgsType then do
+              performEtaExpansion e0 es etaNames etaArgsType
+            else do
+              -- There is at least a non monomorphized type variables
```
facundominguez: The reason to require monomorphized type variables is unknown to me. Are there any conjectures?

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -830,6 +857,81 @@ evalApp γ ctx e0 es et
...
+    flatten t = [t]
+
+    performEtaExpansion :: Expr -> [Expr] -> [Symbol] -> [Sort] -> EvalST (Maybe Expr, FinalExpand)
+    performEtaExpansion fun actualArgs etaNames etaArgsType = do
+      let etaVars = zipWith (\name ty -> ECst (EVar name) ty) etaNames etaArgsType
+      let fullBody = eApps fun (actualArgs ++ etaVars)
+      let etaExpandedTerm = mkLams fullBody (zip etaNames etaArgsType)
+
+      modify $ \st ->
+        st { evNewEqualities = S.insert (eApps fun es, etaExpandedTerm) (evNewEqualities st) }
```
facundominguez: Shouldn't this be skipped when `etaNames` is empty?

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -603,9 +619,20 @@ eval γ ctx et = go
...
+                                    -- we need this to not end in an infinite
+                                    -- loop in which we are eta expaning forever
+                                    -- See comment **ETA**, if we are not performing
+                                    -- any eta expansion then the two branches are actually
+                                    -- equivalent
+                                    if not isLHSApp then do
+                                      (me', fe) <- evalApp γ ctx e [] et
+                                      return (Mb.fromMaybe e me', fe)
+                                    else do
+                                      return (e, noExpand)
```
facundominguez: Only to get more room for the comment:
```suggestion
      -- If it not on the LHS of a function application then expand the
      -- definition we need this to not end in an infinite loop in which
      -- we are eta expanding forever. See comment **ETA**, if we are not
      -- performing any eta expansion then the two branches are actually
      -- equivalent
      if not isLHSApp then do
        (me', fe) <- evalApp γ ctx e [] et
        return (Mb.fromMaybe e me', fe)
      else do
        return (e, noExpand)
```

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -603,9 +619,20 @@ eval γ ctx et = go
...
     go (POr es)         = efAll POr (go `traverse` es)
     go e | EVar _ <- dropECst e = do
-      (me', fe) <- evalApp γ ctx e [] et
-      return (Mb.fromMaybe e me', fe)
-    go (ECst e t)       = do (e', fe) <- eval γ ctx et e
+                                    -- If it not on the LHS of a function
+                                    -- application then expand the definition
+                                    -- we need this to not end in an infinite
+                                    -- loop in which we are eta expaning forever
+                                    -- See comment **ETA**, if we are not performing
```
facundominguez: The comment **ETA** is rather helpful. Can we also have an word about how the infinite loop would unfold here if not for `isLHSApp`. I still have difficulties to see it.

src/Language/Fixpoint/Solver/Common.hs
facundominguez: That looks good enough. Thanks.

facundominguez (commented):
Thanks @AlecsFerra. I added some comments about points I don't understand yet, but your comments helped me progress with my initial questions. Other than these, I guess only some tests would be necessary.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -830,6 +857,81 @@ evalApp γ ctx e0 es et
...
+-- And now PLE can perform the unfolding of f as it's now fully saturated
+-- { \y:Int -> y = something }
+-- Clearly we need the higerorder flag active as we are generating lambda in
+-- the constraints.
+evalApp γ _ctx e0 es _et
+  | EVar f <- dropECst e0
+  , Just eq <- Map.lookup f (knAms γ)
+  = do
+    isHOAllowed <- gets hoFlag
+    if isHOAllowed then do
```
facundominguez: The slowdown in `ple/pos/Permutations.hs` makes me wonder if extensionality shouldn't be required here as well.

Also, the context in which `f` appears seems to be relevant. Eta expansion makes sense in `C f == C g`, or `map f xs == map g xs`, but not in `map f [] == []`. The size of the PLE dumps could be growing when eta expanding unconditionally. But I don't know if there is a good criteria to recognize the relevant contexts.

In the absence of such criteria, I think it might be preferable to have a flag to turn it on.

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: It wont happen, eta name is empty only if the function is fully applied and that case is already covered by the first case of `evalApp`

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: If i put there a type variable then LF is going to complain at some point that it can't unify the type variable with the actual type, you can trigger this error using this code https://gist.github.com/AlecsFerra/c31162353b21e034dc3797c5c227c049 replacing the type of `id` with `a -> a` and removing the check at line 899 inside PLE

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: :+1: Good point. It would deserve adding a comment.

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Sus meant suspicious, it' because at the beginning I tough that I was removing some `ECst` from some variables in some way but actually later I found out that actually they are already generated without the ECst in the previous passes

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: I'll try to explain here and if its ok I'll add as a comment

suppose we have a function f with one argument

We start with the expression `f` when eval is called isLHS is False.
So we end up in the case that is calling eval app on f and we get back `\x -> f x`

Now eval is called on the lambda thus we add in scope x and we call eval on `f x`
We ed up in the case for function application, but before calling evalApp we first call eval on the function and the arguments.

So we end up calling eval on `f` again if we don't have the boolean to tell us that we are on the LHS of a function application we would call again evalApp on `f` alone and obtain again `\x -> f x` then we evaluate the lambda ad so on...

Is it clearer?

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Should we hide it behind the extensionality flag or make a new one? I agree with your point but I can't think of a way to check if the eta expansion is actually needed or not

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: I think that I can remove that comment since its not my fault

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: > Should we hide it behind the extensionality flag or make a new one?

There are only a few tests that enable both extensionality and PLE.
```
tests/import/client/ReflectClient8.hs
tests/neg/T1577.hs
tests/pos/ReWrite2.hs
tests/pos/ReWrite7.hs
tests/pos/ReWrite8.hs
tests/pos/T1560.hs
tests/pos/T1560B.hs
tests/pos/T1577.hs
```

In these cases, eta expansion was not necessary. We have to guess if the authors of these tests would be bothered by the larger PLE dumps. If they are bothered, use a new flag. If they aren't bothered, use extensionality instead.

At this point, I think you and Niki are in a better position to decide. I'm no user of extensionality yet, though I guess the `ple/pos/Permutations.hs` (which I wrote) could be simplified if the support for lambdas is improved.

AlecsFerra: The thing is that we get the eta expansion working 100% (will require some work on lambdas) then everything that is done by extensionality can be done by the eta expansion alone

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: That helps. I'm still grappling with the argument, though.

1. `f` is expanded to `\eta1 -> f eta1`. The next time `eval` is called, it is going to find that `f` is applied to all the expected arguments, and therefore `f eta1` is not going to be eta expanded again. Still looks like `isLHSApp` is not needed (?).

2. If `isLHSApp` is kept, `f :: Int -> Int -> Int` is not going to be eta-expanded in
```
C (f x) == C g
```
because it appears in the function position of `f x`. Is this intended?

AlecsFerra: No look at the code that is evaluating the EApp

```haskell
      go e@(EApp _ _)     =
      case splitEAppThroughECst e of
       (f, es) | et == RWNormal ->
          -- Just evaluate the arguments first, to give rewriting a chance to step in
          -- if necessary
          do
            (es', finalExpand) <- feSeq <$> mapM (eval γ ctx et False) es
            if es /= es'
              then return (eApps f es', finalExpand)
              else do
                (f', fe)  <- eval γ ctx et True f
                (me', fe') <- evalApp γ ctx f' es et
                return (Mb.fromMaybe (eApps f' es') me', fe <|> fe')
       (f, es) ->
          do
            (f', fe1) <- eval γ ctx et True f
            (es', fe2) <- feSeq <$> mapM (eval γ ctx et False) es
            let fe = fe1 <|> fe2
            (me', fe') <- evalApp γ ctx f' es' et
            return (Mb.fromMaybe (eApps f' es') me', fe <|> fe')

```

In all the cases before calling evalApp we are calling eval on f by itself

facundominguez: Oh, regarding (1), I think I understand now why you get into a loop before calling `evalApp`. And regarding (2), I see that it is going to be eta expanded anyway, because `evalApp` will be called on it.

:thinking:

AlecsFerra: Exactly we get
(\y -> f x) y
Now when we evaluate the lambda inside we can unfold f x
then we perform the replacing of x with y in the expansion

AlecsFerra: There is also the possibility that we don't have the implementation of f but it should be the same thing minus the unfolding

facundominguez: How about changing the `EApp` case of `eval` like so
```Haskell
  ...
  (f', fe)  <- case dropECst f of
    EVar _ -> return (f, noExpand)
    _ -> eval γ ctx et True f
  (me', fe') <- evalApp γ ctx f' es et
  ...
```
?

This makes the context of the `EVar` handling more explicit and avoids growing the parameters of `eval`.

AlecsFerra: I don't know if it going to solve the problem in general maybe on the lhs we can end up having something like (id f), I remember that I thought about that solution but then I convinced my self it was not enough but I can easily be wrong

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -829,13 +829,10 @@ evalApp γ ctx e0 es et
...
-                        evPendingUnfoldings = M.insert (eApps e0 es) e3' (evPendingUnfoldings st)
-                      }
-                    -- Don't unfold the expression if there is an if-then-else
-                    -- guarding it, just to preserve the size of further
-                    -- rewrites.
-                    return (Nothing, noExpand)
+                      st { evPendingUnfoldings = M.insert (eApps e0 es) e3' (evPendingUnfoldings st)
+                         , evNewEqualities = S.insert (eApps e0 es, e3') (evNewEqualities st)
+                         }
+                    return (Just $ eApps e2' es2, fe)
```
facundominguez: :wave: It is fine to experiment, but this kind of change is going to need special consideration as it grows the amount of potentially irrelevant equations.

AlecsFerra: Yeah my bad I meant to push it on the other branch, anyway when I'm done I'm gonna hide it under a flag

AlecsFerra:
Still in todo:
- ~~Check if all the equations are actually needed~~
- Find a more solid way to write `validVar' [here](https://github.com/AlecsFerra/liquid-fixpoint/blob/6ec234a7811a780e366f15d60d563385c76993b7/src/Language/Fixpoint/Solver/PLE.hs#L341)
- Decide if the feature will be hidden by a flag and if yes which 
- ~~Think if [this](https://github.com/ucsd-progsys/liquid-fixpoint/pull/705#discussion_r1763372545) is enough~~

AlecsFerra:
Ops forgot about the tests on liquid-fixpoint

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: It wasn't actually needed you're right

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Corrected

AlecsFerra:
filtered
top
bot

AlecsFerra:
I can open a PR on LH when we are done here https://github.com/AlecsFerra/liquidhaskell/commits/develop/ I've also added two tests to showcase the feature:
- The famous STLC to SKI proof that started the whole thing [here](https://github.com/AlecsFerra/liquidhaskell/blob/develop/tests/ple/pos/SKIEta.hs)
- Automatic proof of the monad axiom of the reader monad [here](https://github.com/AlecsFerra/liquidhaskell/blob/develop/tests/ple/pos/ReaderEta.hs)

AlecsFerra:
Btw I think that now we can allow lambdas in reflected functions definitions, at least for the reader monad example it works

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -266,9 +271,97 @@ withAssms :: InstEnv a -> ICtx -> Diff -> Maybe SubcId -> (InstEnv a -> ICtx ->
...
+      | EVar yn <- dropECst y
+      , validVar yn
+      = fmap ((yn, x) :) (unifyArgumets xs ys)
+    unifyArgumets (x:xs) (y:ys)
+      | dropECst x == dropECst y
+      = unifyArgumets xs ys
+    unifyArgumets _ _ 
+      = Nothing
+
+    extractRewrites (_, sreft, _)
```
facundominguez: The purpose of this function needs to be documented. It is still unreachable to me.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -266,9 +271,97 @@ withAssms :: InstEnv a -> ICtx -> Diff -> Maybe SubcId -> (InstEnv a -> ICtx ->
...
+    extractRewrites _ = []
+
+    keepPropEq (PAtom Eq e1 e2) = Just (e1, e2)
+    keepPropEq _ = Nothing
+    
+    keepTrans [a, b] = Just (a, b)
+    keepTrans _ = Nothing
+
+-- Find a way to make this check more roboust
+-- Check if a symbol is a local vairable
```
```suggestion
-- Find a way to make this check more robust
-- Check if a symbol is a local variable
```

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -266,9 +271,97 @@ withAssms :: InstEnv a -> ICtx -> Diff -> Maybe SubcId -> (InstEnv a -> ICtx ->
...
+validVar :: Symbol -> Bool
+validVar name =  not (T.elem '.' name')
+              && (T.elem '#' name' || T.elem '_' name')
+              && not (T.isPrefixOf "lq_anf" name')
+  where name' = symbolText name
```
facundominguez: What does it mean for a variable to be local in liquid-fixpoint? I could understand the concept in Haskell, but not so sure in this context.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -367,12 +460,13 @@ data InstEnv a = InstEnv
...
-  , icSimpl    :: !ConstMap                 -- ^ Map of expressions to constants
-  , icSubcId   :: Maybe SubcId              -- ^ Current subconstraint ID
-  , icANFs     :: [[(Symbol, SortedReft)]]  -- Hopefully contain only ANF things
+  { icAssms       :: S.HashSet Pred           -- ^ Equalities converted to SMT format
+  , icCands       :: S.HashSet Expr           -- ^ "Candidates" for unfolding
+  , icEquals      :: EvEqualities             -- ^ Accumulated equalities
+  , icSimpl       :: !ConstMap                -- ^ Map of expressions to constants
+  , icSubcId      :: Maybe SubcId             -- ^ Current subconstraint ID
+  , icANFs        :: [[(Symbol, SortedReft)]] -- Hopefully contain only ANF things
+  , icSMTRewrites :: M.HashMap Symbol Expr    -- ^ Rewrites that have been generated by SMT
```
facundominguez: What is a rewrite, and how does the SMT solver generate it? Note that we must be able to distinguish these from REST rewrites and equalities generated by PLE.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -625,22 +733,54 @@ evalELam :: Knowledge -> ICtx -> EvalType -> (Symbol, Sort) -> Expr -> EvalST (E
...
 data RESTParams oc = RP
   { oc   :: OCAlgebra oc Expr IO
   , path :: [(Expr, TermOrigin)]
   , c    :: oc
   }
 
+-- With the support for lambdas in PLE now lambda expression can interact with the 
+-- rewriting feature. But for the moment we don't support them as we can't convert to
+-- REST terms.
+isExprRewritable :: Expr -> Bool
```
facundominguez, line 671: Please, document what it means for an expression to be rewritable.

Also the comment above this function should be deleted, or it should go into the trouble of explaining how a lambda expression can interact with REST. It should also make explicit why lambdas cannot be converted to REST terms (I'm guessing they just do not offer a way to express them).

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -580,20 +681,26 @@ eval γ ctx et = go
...
+                  _      -> eval γ ctx et f
                 (me', fe') <- evalApp γ ctx f' es et
                 return (Mb.fromMaybe (eApps f' es') me', fe <|> fe')
        (f, es) ->
           do
-            (f':es', fe) <- feSeq <$> mapM (eval γ ctx et) (f:es)
+            (f', fe1) <- case dropECst f of
+              EVar _ -> pure (f, noExpand)
+              _      -> eval γ ctx et f
+            (es', fe2) <- feSeq <$> mapM (eval γ ctx et) es
```
facundominguez, line 604: You might want to use a `go` call in all places here where `eval` is called.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -625,22 +733,54 @@ evalELam :: Knowledge -> ICtx -> EvalType -> (Symbol, Sort) -> Expr -> EvalST (E
 evalELam γ ctx et (x, s) e = do
     oldPendingUnfoldings <- gets evPendingUnfoldings
     oldEqs <- gets evNewEqualities
+
+    -- We need to declare the variable in the environment 
+    modify $ \st -> st 
+      { evEnv = declareVar x s $ evEnv st }
```
facundominguez: Use `insertSortEnv` instead?

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -625,22 +733,54 @@ evalELam :: Knowledge -> ICtx -> EvalType -> (Symbol, Sort) -> Expr -> EvalST (E
...
+
     (e', fe) <- eval (γ { knLams = (x, s) : knLams γ }) ctx et e
     let e2' = simplify γ ctx e'
         elam = ELam (x, s) e
     -- Discard the old equalities which miss the lambda binding
     modify $ \st -> st
       { evPendingUnfoldings = oldPendingUnfoldings
       , evNewEqualities = S.insert (elam, ELam (x, s) e2') oldEqs
+      -- Leaving the scope thus we need to get rid of it
+      , evEnv = deleteVar x $ evEnv st
```
facundominguez: Either define `deleteSymEnv` next to `insertSymEnv`, or just pass here the env before the insertion.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -786,38 +933,31 @@ evalApp γ ctx e0 es et
   , Just eq <- Map.lookup f (knAms γ)
   , length (eqArgs eq) <= length es
   = do
-       env  <- gets (seSort . evEnv)
+       env <- gets (seSort . evEnv)
```
facundominguez: Trailing whitespace.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -831,6 +971,9 @@ evalApp γ ctx e0 es et
       = arg
       | otherwise = e
 
+    isGuardUndecied EIte{} = True
+    isGuardUndecied _ = False
```
facundominguez: Maybe rename to `hasUndecidedGuards`. Note the typo in the current name.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -875,13 +1018,80 @@ evalApp γ ctx e0 es et
         else do
           return (Nothing, noExpand)
 
+evalApp _ ctx e0 es _
+  | EVar f <- dropECst e0
+  , Just rewrite <- M.lookup f (icSMTRewrites ctx)
+  = do
+    -- No need to check if the eta beta flag is on since icSMTRewrites is emoty
```
facundominguez: It is not apparent to me why `icSMTRewrites` is empty at this point. Also, note the typo in the comment.

Also I don't see the relevance of emptiness for the case at hand.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -875,13 +1018,80 @@ evalApp γ ctx e0 es et
...
+    modify $ \st ->
+      st { evNewEqualities = S.insert (eApps e0 es, eApps rewrite es) (evNewEqualities st) }
+    return (Just $ eApps rewrite es, expand)
+
+evalApp _ ctx e0 es _
+  | deANFed <- deANF ctx e0
```
facundominguez: More idiomatic
```suggestion
  | let deANFed = deANF ctx e0
```

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -875,13 +1018,80 @@ evalApp γ ctx e0 es et
...
+  , dropECst deANFed /= dropECst e0
+  = do
```
facundominguez: The need to inline ANF bindings needs to be documented. I don't get it yet.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -875,13 +1018,80 @@ evalApp γ ctx e0 es et
...
+      if isEtaBetaOn then
+        return (Just $ eApps deANFed es, expand)
+      else
+        return (Nothing, noExpand)
+
+-- **ETA**: PLE before the patch to perform eta expansion was only able to unfold
+-- fully applied functions, by eta equivalence we mean the property that for all
+-- functions f the terms f and (\x -> f x) have the same semantics, we leverage 
+-- this fact to fully apply functions with extra function arguments so to trigger
+-- the standard PLE espansion, see the following example:
```
```suggestion
-- the standard PLE expansion, see the following example:
```

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -875,13 +1018,80 @@ evalApp γ ctx e0 es et
...
+      -- only way for etaNames to be empty is if the function is fully applied, but that case
+      -- is already handled by the previous case of evalApp
+      modify $ \st -> st 
+        { evNewEqualities = S.insert (eApps e0 es, etaExpandedTerm) (evNewEqualities st) }
+      return (Just etaExpandedTerm, expand)
+    else do
+      pure (Nothing, noExpand)
+  where
+    flatten (FFunc t ts) = t : flatten ts
+    flatten t = [t]
```
facundominguez: How about
```Haskell
    unpackFFuncs :: Sort -> ([Sort], Sort)
    unpackFFuncs (FFunc t ts) =
      let (args, ret) = unpackFFuncs ts
       in (t : args, ret)
    unpackFFuncs t = ([], t)
```
It would avoid the init and drop calls above.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -786,38 +933,31 @@ evalApp γ ctx e0 es et
...
-                  else do
-                    useFuel f
-                    modify $ \st ->
-                      st
-                        { evNewEqualities = S.insert (eApps e0 es, e3') (evNewEqualities st)
-                        , evPendingUnfoldings = M.delete (eApps e0 es) (evPendingUnfoldings st)
-                        }
-                    return (Just $ eApps e2' es2, fe)
-         else return (Nothing, noExpand)
+       if okFuel && et /= FuncNormal then do
```
facundominguez, line 885: Indentation should be at least two spaces in the branches of the `if`.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -786,38 +933,31 @@ evalApp γ ctx e0 es et
...
+        let (es1, es2) = splitAt (length (eqArgs eq)) es
+        -- Doing elaboration ahead of time would cause too much
+        -- monomorphization
```
facundominguez: I'm guessing every reader of this comment will want to know:
1. for what purpose monomorphization would be too much,
2. what does monomorphization mean in this context
3. what is the interaction between elaboration and whatever purpose is claimed in (1).

facundominguez (commented):
Thanks @AlecsFerra.

I did a first pass and left mostly requests for documentation. I still need to understand some parts.

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Local in the sense that the argument of a function ex:

```haskell
{-@ f :: a:Int -> b:Int -> { a + b = b + a } @-}
```

`a` and `b` are local

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: When we perform pattern matching some quality of the shape explained at line 74 are inserted, the rewrites extracted from there are inserted in icSMTRewrites

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: I can't find that function

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: I see. But I'm afraid that it is a concept from liquidhaskell, liquid-fixpoint has no notion of function specification, does it?

We have instead an environment of bindings, subtyping constraints, and function definitions. Could we explain what PLE is supposed to do, if we lack the concept of function specs?

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Excuse me, I provided the wrong name. Here it is:
```
src/Language/Fixpoint/Types/Theories.hs:insertSymEnv :: Symbol -> Sort -> SymEnv -> SymEnv
```

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: You can see that at line 275 that if the flag is not enabled we return an empty list.
The comment is justifying why I'm not checking if the etabeta flag is enabled

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Didn't know that was allowed cool

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Sometimes we have stuff to expand that is in the definition of the anf

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Thanks. I'm deferring this conversation until I understand better the need for `icSMTRewrites`.

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: I think the delete function is better as it's better at communicating the intent

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: The core difference is that top level definitions types ecc have a definition, "locals" don't have one so you can infer their definition during unification

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: I added a comment lmk if it helps, this function is doing part of the work explained in the comment above obtainEqualities

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: It means that it can be converted to a runtime term of the rest library, lambdas are unsupported I don't know how to make the comment more understandable

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Sometimes haskell type variables are treated differently

for example in the callsite for the id function sometimes is annotated with liquid type variables (func(0, [@12, @12]) and sometimes they get monomorphized to something like
func(0, [a#12312, a#12312]) and during eta expansion will cause typing problems

AlecsFerra:
I will squash the commits at the end

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -266,9 +271,99 @@ withAssms :: InstEnv a -> ICtx -> Diff -> Maybe SubcId -> (InstEnv a -> ICtx ->
...
+      , validVar yn
+      = fmap ((yn, x) :) (unifyArgumets xs ys)
+    unifyArgumets (x:xs) (y:ys)
+      | dropECst x == dropECst y
+      = unifyArgumets xs ys
+    unifyArgumets _ _ 
+      = Nothing
+
+    -- Looks inside a list of qualities for equalities 
+    -- that share the lhs and then perform the unification
```
facundominguez: A more accurate description could be:
```
    -- Yields the unification of the equalities present in the refinement
    -- type of a binding
```
But this documentation still gives no clue of why someone would care of constructing this unification.

Moreover, it is looking to me like the equalities that unification provides are redundant. That is, the SMT solver can already infer that they hold from the equalities present in the binding (which are fed as assumptions).

Maybe if these were used with `--no-adts` in LH, then it would make more of a difference.

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: As I see it, these equalities do not define the variables. Instead, they assert additional assumptions about them, whether the variable is local doesn't seem to invalidate them.

Do you have an example where it goes wrong for non-local variables?

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: That answers (2) well, and (1) has something to do with eta-expansion. 

> ... and during eta expansion will cause typing problems

What typing problems are these?

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Yeah the SMT knows them but ple doesn't

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: You can remove the check and try SKIEta.hs

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: When you perform the replacement the equation before being sent to the SMT solver will be elaborated and it will fail since the unification will fail

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: > I don't know how to make the comment more understandable

I think the comment should make explicit which expressions are considered to be rewritable/convertible-to-rest, and which expressions are considered not to be. Then we can inspect the implementation to know if it is faithful to the intent.

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Following this in another conversation.

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Ah, can we have the comment explain why PLE needs to unfold those?

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Thanks, though that's going to be some debugging exercise. If you already understand the interaction, you could try putting your account of it somewhere for the sake of the review.

AlecsFerra: Yeah in a comment of here? The intuition is imaginenyou have to unify `local_variable = real_function`
You would like only the rewrite `local_variable` -> `real_function` and not the opposite as other than being fairly useless it would lead to an infinite loop

facundominguez: > Yeah in a comment of here?

Eventually in a comment, but either works for now.

> and not the opposite as other than being fairly useless it would lead to an infinite loop

That's helpful to know. Which leads me back to figuring out why PLE needs to apply these rewrites to begin with before continuing.

AlecsFerra: Yeah lets take as an example this [theorem](https://github.com/AlecsFerra/liquidhaskell/blob/29ee6e1a524b195bfb97bcd0a9550f497aba0c2a/tests/ple/pos/SKIEta.hs#L239) here knowing what f is equal to allows us to perform more unfoldings and prove more things, let's focus on this [case](https://github.com/AlecsFerra/liquidhaskell/blob/29ee6e1a524b195bfb97bcd0a9550f497aba0c2a/tests/ple/pos/SKIEta.hs#L262) in this specific case by unification we obtain that `f = lookupRef ... (Here ...)` so PLE now will unfold `lookupRef` and so on... genertaing the various qualities (as we have a specific instance/call of `lookupRef`), instead if we don't perform the rewrite yes the SMT will still know that `f = lookupRef ... (Here ...)` but it won't unfold the call to lookup red and subsequently it won't know all the other equalities

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: I've looked [here](https://github.com/ucsd-progsys/liquid-fixpoint/blob/4233197fbefa143537d8dc17f6a21be5d88f5d57/src/Language/Fixpoint/Solver/Rewrite.hs#L92), I honestly don't understand fully the REST code

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Responded in the other thread

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: I think that theorem is too complicated. There must be a short lemma that demonstrates the need, it doesn't have to come from an LH example necessarily. But let's grapple with yours for the sake of understanding your point. I have difficulties to prove it.

### Environment:
```
r = Here _ _
e = CVar _ _ r
prop r = Ref σ0 (Cons σ0 γ0)            // From type inference
prop e = Comb γ1 σ0 (lookupRef σ0 γ1 r) // From type inference
prop r = Ref σ0 γ1                      // From type inference
prop e = Comb (Cons σ γ) τ f            // From spec of bracket

f = lookupRef σ0 γ1 r                   // unifying "prop e" equalities

v = I σ γ                                // definition of bracket
prop v = Comb γ (iType σ) (makeId σ γ)   // From type inference
```

### Function definitions

```
makeId σ γ v = (VFun σ σ id)
id a = a
iType σ = Arrow σ σ
lookupRef _ _ (Here _ _) (With _ _ a _) = a
makeBracketing σ τ γ f env = VFun σ τ (makeBracketingInside σ τ γ f env)
makeBracketingInside σ τ γ f env x = f (With σ γ x env)
```

### Subtyping constraint:

```
prop v = Comb γ (iType σ) (makeId σ γ)
==>
prop v = Comb γ (Arrow σ τ) (makeBracketing σ τ γ f)
```

### Proof

```
prop v = Comb γ (iType σ) (makeId σ γ)
 ==>   // Definition of iType
prop v = Comb γ (Arrow σ σ) (makeId σ γ)
 ==>   // proof obligation: 1) σ = τ && 2) makeId σ γ = makeBracketing σ τ γ f
prop v = Comb γ (Arrow σ τ) (makeBracketing σ τ γ f)
```

#### Proof obligation (1): σ = τ 

Proved by the SMT solver by the equalities on `prop r` and `prop e`

#### Proof obligation (2): makeId σ γ = makeBracketing σ τ γ f

```
makeBracketing σ τ γ f
==>   // definition of makeBracketing and eta expansion
\eta1 -> VFun σ τ (makeBracketingInside σ τ γ f eta1)
==>   // Proof obligation (1)
\eta1 -> VFun σ σ (makeBracketingInside σ τ γ f eta1)
==>   // definition of makeBracketingInside
\eta1 -> VFun σ σ (\eta2 -> f (With σ γ eta2 eta1))
==>   // unification of f with lookupRef
\eta1 -> VFun σ σ (\eta2 -> lookupRef σ0 γ1 r (With σ γ eta2 eta1))
==>   // definition of lookupRef
\eta1 -> VFun σ σ (\eta2 -> eta2)
==>   // definition of id, alpha renaming, and eta conversion
\eta1 -> VFun σ σ id
==>   // definition of makeId, alpha renaming, and eta conversion
makeId σ γ
```
Update: the proof is now complete after further discussion.

AlecsFerra: if you know that `f = lookupRef σ (Cons σ γ) (Here ...)` then you perform the substitution in the equation and you get
```
\eta1 -> VFun σ σ (lookupRef σ (Cons σ γ) (Here ...) (With σ γ x eta1))
==> // Definition of lookupRef
VFun σ σ (\eta1 -> \eta1)
```
The alpha renaming in not the job of PLE, it's done for free by the SMT

Do you think that [this](https://github.com/AlecsFerra/liquidhaskell/blob/29ee6e1a524b195bfb97bcd0a9550f497aba0c2a/tests/ple/pos/ReaderEta.hs#L46) is better? I cited the SKI example because it was the one that started everything

facundominguez: > if you know that f = lookupRef σ (Cons σ γ) (Here ...)

Oh, I got it now. I updated the proof. Will think about it some more.

> Do you think that this is better?

Maybe. We would have to elaborate the proof to see.

facundominguez: Here's a trimmed down example that could work for explanation, together with references to your tests:
```
definitions
g x y = y
id x = x

Environment:
v = C g
v = C f

Constraint:
D (f x) ==> D id

Proof:

D (f x)
==>   // unification: (C g = C f ==> f = g)
D (g x)
==>   // eta expansion
D (\eta1 -> g x eta1)
==>   // definition of g
D (\eta1 -> eta1)
==>   // definition of id
D id
```

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -266,9 +271,99 @@ withAssms :: InstEnv a -> ICtx -> Diff -> Maybe SubcId -> (InstEnv a -> ICtx ->
...
+
+    keepPropEq (PAtom Eq e1 e2) = Just (e1, e2)
+    keepPropEq _ = Nothing
+    
+    keepTrans [a, b] = Just (a, b)
+    keepTrans _ = Nothing
+
+-- Find a way to make this check more robust
+-- Check if a symbol is a local variable
+validVar :: Symbol -> Bool
```
facundominguez: My account of what this implements: users of LF are asked to mark variables that are considered for unification-induced rewriting by giving them special names (you called them local so far). The users are responsible for avoiding non-confluent rewriting rule sets that could result from these unifications.

It looks like a hard proposition. Firstly, predicting the rule sets that will be produced requires calculation. And second, determining termination is non-trivial.

A partial solution would be to do the unification in LH, and augment the fq file/info with rewrites that apply per constraint rather than globally. This solution is partial because it makes explicit the rewrites (good), but it still leaves to the user to ensure termination. On the other hand, maybe this isn't worse than asking the user to ensure termination of reflected functions.

Lastly, is requiring local variables on the left of a rewrite `(x, e)` sufficient to ensure termination? Maybe `e` should not contain local variables, or some other restriction of the sort.

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Ok, but there is no need to understand REST. What we would need is a statement like, e.g. "an expression is rewritable if it contains no lambdas."

Then we can check that the implementation correctly flags all expressions containing at least one lambda.

Then someone who understands REST can evaluate one day if this definition of rewritable is overly restrictive.

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: I think you can't have a constraint that isn't false that is cyclic since they are required to be the argument of a contractor, you're right about the naming thing but I couldn't find any other mechanisms, I would have expected variables to be tagged differently based on the context they are introduced like in ghc

AlecsFerra: Another solution that maybe it's also retro compatible is having a list of all the definitions like we have with type constructors

AlecsFerra: Then a local variable is anything that is not inside the definitions list && is not inside the ANF list

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: I don't think I follow your explanation. Could you make an example to illustrate perhaps?

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: ... and why does the ANF binding need to be inlined for the "stuff" to be expanded?

I could imagine `let anf = f x in anf y` being inlined for `f x y` to be exposed, but this explanation still doesn't let me discern in which cases it would appear, or why it is a problem only for lambdas. Could it be made more specific?

AlecsFerra: Because from what I understood the anf were supposed to be substituted before executing eval, but probably this isn't enough when performing substitutions obtained through unification

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Oh well, I guess it could also say something like: "an expression is rewritable if it is in the domain of Language.Fixpoint.Solver.Rewrite.convert". Then the documentation of `convert` would need to refer back to `isExprRewritable` to keep both functions in sync.

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: That looks like a possible mechanism to mark variables. But let me stress that the user experience problem I'm surfacing is rather independent of how the local variables are communicated to LF.

AlecsFerra: I don't really understand what is the problem then

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: If you have an equation for id if you run unify on it's body (meaning on the field eqBody of the Equation) you will get all the variables annotates with the monomorphized type `a#foo` and not with the liquid type variable `@foo` so when you perform the substitution in some place where the type variable is instantiated to some concrete type 
Bar, the evalCandsLoop will typecheck and complain that it cannot unify `a#foo` with `Bar` instead if we do it later and we can keep the liquid type variable during typechecking `@foo` is unified with `Bar`

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: > ... is requiring local variables on the left of a rewrite (x, e) sufficient to ensure termination? Maybe e should not contain local variables, or some other restriction of the sort.

Here's an example I am imagining:

```Haskell
data C = C (Int -> Int)

{-@ lemma :: g:(Int -> Int) -> {f:(Int -> Int) | C f = C g && C g = C f} -> { f = g } @-}
lemma :: (Int -> Int) -> (Int -> Int) -> ()
lemma _ _ = ()
```

I would expect unification to produce rewrites `(f, g)` and `(g, f)` in this case.

AlecsFerra: This code wont trigger PLE but it can be modified to do so:

```haskell
{-@ LIQUID "--reflection"                @-}
{-@ LIQUID "--ple"                       @-}
{-@ LIQUID "--etabeta"                   @-}
{-@ LIQUID "--save"                      @-}

module Counter where

import Prelude hiding (id, const)

data C = C (Int -> Int)

{-@ reflect id @-}
id :: a -> a
id x = x

{-@ reflect const @-}
const :: a -> b -> a
const x _ = x

{-@ h :: g:_ -> {f:_ | C f = C g && C g = C f} -> {k:_ | k = C f} -> {q:_ | q = C g} -> { f = const id g } @-}
h :: (Int -> Int) -> (Int -> Int) -> C -> C -> ()
h _ _ (C f) (C g) = ()
```

2 things are required to trigger my code:
- Trigger PLE thus we need a non trivial goal for the SMT
- Case splitting to generate the rewrites

Those 2 case split put information in two different bind hence we can't end in an infinite loop

AlecsFerra: You can look at the rewrites by adding by adding
```haskell
  traceShowM ("ID" :: String, cidMb)
  traceShowM ("REWRITES: " :: String, pprint smtRewrites)
```
At line 281 in `PLE.hs`

facundominguez: The unification is very sensitive to the way in which the equalities are written. Here's our infinite loop.
```Haskell
{-@ LIQUID "--ple" @-}
{-@ LIQUID "--reflection" @-}
{-@ LIQUID "--extensionality" @-}
{-@ LIQUID "--etabeta" @-}

module Test where

data C = C (Int -> Int)

{-@
lemma
  :: k:C
  -> l:C
  -> g:(Int -> Int)
  -> f:(Int -> Int)
  -> { v:()
     |    k = C f
       && k = C (\x:Int -> g x)
       && l = C g
       && l = C (\y:Int -> f y)
     }
  -> { f = g }
@-}
lemma :: C -> C -> (Int -> Int) -> (Int -> Int) -> () -> ()
lemma _ _ _ _ _ = ()
```

AlecsFerra: Um you're right probably we can detect loops by looking at the rewrites but it would also make us lose some rewrites that are ok

AlecsFerra: But probably before we should decide what to do with variables

AlecsFerra: After thinking a bit abou that I underythe issue those rewrites do not come by a case split

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: ANF inlining is needed for REST, but it is not needed to unfold functions, and that's why it was not called in the rest of PLE.

I still don't know what this case is needed for.

AlecsFerra:
Following some discussion in private I decided to remove the code relating to make PLE aware of dependent pattern matching / unification

The code of the reader monad still passes, the SKI calculus instead will require some extra lemmas

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Deferred by removal of the change.

facundominguez: Deferred by removal of the change.

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Deferred by removal of the change.

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: Deferred by removal of the change.

facundominguez:
The only open discussion that I could catch is about making a comment [more instructive](https://github.com/ucsd-progsys/liquid-fixpoint/pull/705#discussion_r1782705115). Update: oh, I asked about a couple other things in comments later.

Regarding performance, if the verification times worsen when enabling extensionality and `--etabeta` I'd say it is fine to merge. If verification worsens even when not enabling the flags, then I'd prefer to understand why before merging.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -786,38 +836,31 @@ evalApp γ ctx e0 es et
...
-
-                if undecidedGuards
-                  then do
-                    modify $ \st ->
-                      st {
-                        evPendingUnfoldings = M.insert (eApps e0 es) e3' (evPendingUnfoldings st)
-                      }
-                    -- Don't unfold the expression if there is an if-then-else
-                    -- guarding it, just to preserve the size of further
-                    -- rewrites.
```
facundominguez, line 810: This comment was lost when reformatting.

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -875,13 +921,61 @@ evalApp γ ctx e0 es et
...
+-- { f = something }
+-- The original PLE cannot do anything on f as it's not fully applied, instead
+-- if we perform eta expansion on f we obtain the following constraint
+-- { \y:Int -> f y = something }
+-- And now PLE can perform the unfolding of f as it's now fully saturated
+-- { \y:Int -> y = something }
+-- Clearly we need the higerorder flag active as we are generating lambda in
+-- the constraints.
+evalApp _γ _ctx e0 es _et
+  | ECst (EVar _f) sortAnnotation@FFunc{} <- e0
```
facundominguez, line 987: I just noticed this is relying on `ECst` while the earlier equations look for the function definition in `γ`. What is the reason to not do the same here?

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: In addition, the comment should probably explain why elaborating the term is necessary at this point.

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: I've added a comment in the code

AlecsFerra:
filtered
bot
top

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -786,38 +836,42 @@ evalApp γ ctx e0 es et
...
+         -- are needed for doing eta expansion, doing elaboration
+         -- of the function body ahead of time would cause too much
+         -- monomorphization on the LH side, suppose we have a
+         -- polymorphic function of Haskell type `forall a. (a -> a)`,
+         -- we want the type `forall a. (ECst a a -> ECst a a)`
+         -- if we perform the elaboration ahead of time we would
+         -- get a term annoptated with type `fun(0, [a#foo, a#foo])`
+         -- instead of `fun(0, [@foo, @foo])` thus when substituting
+         -- the body in a term we could encounter type mismatches.
+         newE <- elaborateExpr "EvalApp unfold full: " $ substEq env eq es1
```
facundominguez: If this had an effect on the performance of the tests that are slowing down, perhaps elaboration could be done only if `--etabeta` is enabled.

facundominguez: The result of the elaboration is going both into the new equalities and into the rewritten term that is returned. I think only the later is necessary, as it is the only place where `ECst` will be looked for by `evalApp`.

AlecsFerra:
filtered
bot
top

AlecsFerra:
Looks way better I guess the slowdowns were actually caused by the elaboration

facundominguez (approved):
@AlecsFerra, would you like to squash or make other edits ahead of merging?

facundominguez:
Also, note the comment edits that I just contributed. Feel free to adjust them if you see anything amiss.

AlecsFerra:
Ops

AlecsFerra:
Ok now It's ok to merge

PR #710, opened 2024-10-11
Local rewrites

AlecsFerra:
Following the discussion on #705 the local rewrites features is implemented standalone.

This PR implements proposal 1 of [#2371/liquid-haskell](https://github.com/ucsd-progsys/liquidhaskell/pull/2371)

Depends on [#2384/liquid-haskell](https://github.com/ucsd-progsys/liquidhaskell/pull/2384)

WIP: at the moment we don't read the local rewrites from `horn` files.

### DIFF
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -147,9 +147,7 @@ instEnv cfg info cs restSolver ctx = do
               , evNewEqualities = mempty
               , evSMTCache = mempty
               , evFuel = defFuelCount cfg
-              , extensionalityFlag = extensionality cfg
               , freshEtaNames = 0
-              , etabetaFlag = FC.etabeta cfg
               , explored = Just et
               , restSolver = restSolver
               , restOCA = restOrd
@@ -223,7 +221,9 @@ pleTrie t env = loopT env' ctx0 diff0 Nothing res0 t
       , icSimpl       = mempty
       , icSubcId      = Nothing
       , icANFs        = []
-      , icLRWs         = mempty
+      , icLRWs        = mempty
+      , etaBetaFlag   = etabeta        $ ieCfg env
+      , extFlag       = extensionality $ ieCfg env
       }
 
 loopT
@@ -381,6 +381,11 @@ data ICtx    = ICtx
   , icSubcId      :: Maybe SubcId             -- ^ Current subconstraint ID
   , icANFs        :: [[(Symbol, SortedReft)]] -- Hopefully contain only ANF things
   , icLRWs        :: LocalRewrites            -- ^ Local rewrites
+  , etaBetaFlag   :: Bool                     -- ^ True if the etabeta flag is turned on, needed 
+                                              -- for the eta expansion reasoning as its going to 
+                                              -- generate ho constraints
+                                              -- See Note [Eta expansion].
+  , extFlag       :: Bool                     -- ^ True if the extensionality flag is turned on
   }
 
 ----------------------------------------------------------------------------------------------
@@ -469,15 +474,10 @@ data EvalEnv = EvalEnv
                                     -- an expression
   , evSMTCache :: M.HashMap Expr Bool -- ^ Whether an expression is valid or its negation
   , evFuel     :: FuelCount
-  , extensionalityFlag :: Bool -- ^ True if the extensionality flag is turned on
   
   -- Eta expansion feature
   , freshEtaNames :: Int -- ^ Keeps track of how many names we generated to perform eta 
                          --   expansion, we use this to generate always fresh names
-  , etabetaFlag        :: Bool -- ^ True if the etabeta flag is turned on, needed for the eta
-                               -- expansion reasoning as its going to generate ho constraints
-                               -- See Note [Eta expansion].
-
   -- REST parameters
   , explored   :: Maybe (ExploredTerms RuntimeTerm OCType IO)
   , restSolver :: Maybe SolverHandle
@@ -893,8 +893,9 @@ evalApp γ ctx e0 es et
          let (es1, es2) = splitAt (length (eqArgs eq)) es
          -- See Note [Elaboration for eta expansion].
          let newE = substEq env eq es1
-         isEtaBetaOn <- gets etabetaFlag
-         newE' <- if isEtaBetaOn then elaborateExpr "EvalApp unfold full: " newE else pure newE
+         newE' <- if etaBetaFlag ctx 
+                    then elaborateExpr "EvalApp unfold full: " newE 
+                    else pure newE
 
          (e', fe) <- evalIte γ ctx et newE'        -- TODO:FUEL this is where an "unfolding" happens, CHECK/BUMP counter
          let e2' = stripPLEUnfold e'
@@ -959,11 +960,10 @@ evalApp γ ctx e0 es _et
 evalApp γ ctx e0 es et
   | ELam (argName, _) body <- dropECst e0
   , lambdaArg:remArgs <- es
+  , etaBetaFlag ctx || extFlag ctx
   = do
       isFuelOk <- checkFuel argName
-      isExtensionalityOn <- gets extensionalityFlag
-      isEtaBetaOn <- gets etabetaFlag
-      if isFuelOk && (isExtensionalityOn || isEtaBetaOn)
+      if isFuelOk
         then do
           useFuel argName
           let argSubst = mkSubst [(argName, lambdaArg)]
@@ -993,7 +993,7 @@ evalApp _ ctx e0 es _
       return (Just $ eApps deANFed es, expand)
 
 
-evalApp _γ _ctx e0 es _et
+evalApp _γ ctx e0 es _et
   -- We check the annotation instead of the equations in γ for two reasons.
   --
   -- First, we want to eta expand functions that might not be reflected. Suppose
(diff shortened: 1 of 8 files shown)


src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -969,6 +975,23 @@ evalApp γ ctx e0 es et
...
+      let expandedTerm = eApps rw es
+      modify $ \st -> st 
+        { evNewEqualities = S.insert (eApps e0 es, expandedTerm) (evNewEqualities st) }
+      return (Just expandedTerm, expand)
+
+evalApp _ ctx e0 es _
+  | deANFed <- deANF ctx e0
+  , dropECst deANFed /= dropECst e0
+  = do
+      return (Just $ eApps deANFed es, expand)
```
facundominguez: We would need an example of an expression that requires to inline anf bindings to unfold function calls in order to motivate this case.

facundominguez (commented):
It is looking simple enough to me.

Some tests would be needed.

Also, the PR description says this is not implemented for the fq format, but it looks like it is.

AlecsFerra:
> Also, the PR description says this is not implemented for the fq format, but it looks like it is.

Forgot to update the comment

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: The issue is that I don't have a small example but the SKI example without lemmas doesn't pass without unfolding ANFs

facundominguez: That sounds like we will need to do some debugging to explain why deANF is needed then.

AlecsFerra: I suspect that is some bug in the code that does the deANFing because I suppose that ANF expressions shouldn't reach the evaluation loop, also the only other place where is called is in some code that does the REST rewrites

facundominguez: Yes, REST does need to remove anf bindings, and that use I think I understand.

AlecsFerra: The weird thing is that the anf terms aren't introduced by the local rewrites they are already there, maybe is just the SMT solver that has trouble to see qualities when they are inside a lambda

nikivazou:
There is a parser correct? If so, can you also add a small `.fq` file to see how it works?

nikivazou (approved):

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -377,6 +381,13 @@ data ICtx    = ICtx
   , icSimpl       :: !ConstMap                -- ^ Map of expressions to constants
   , icSubcId      :: Maybe SubcId             -- ^ Current subconstraint ID
   , icANFs        :: [[(Symbol, SortedReft)]] -- Hopefully contain only ANF things
+  , icLRWs        :: LocalRewrites            -- ^ Local rewrites
+  , etaBetaFlag   :: Bool                     -- ^ True if the etabeta flag is turned on, needed 
+                                              -- for the eta expansion reasoning as its going to 
+                                              -- generate ho constraints
+                                              -- See Note [Eta expansion].
+  , extFlag       :: Bool                     -- ^ True if the extensionality flag is turned on
+  , lrwsFlag      :: Bool                     -- ^ True if the locad rewrites flag is turned on
```
```suggestion
  , lrwsFlag      :: Bool                     -- ^ True if the local rewrites flag is turned on
```

facundominguez: It is worth following the naming convention of the other flags here and rename these to `icEtaBetaFlag`, `icExtensionalityFlag`, `icLocalRewritesFlag`.

Also the short forms `extFlag` or `lrwsFlag` are faster to write but unfortunately cryptic for the readers of the code.

AlecsFerra:
> There is a parser correct? If so, can you also add a small `.fq` file to see how it works?

Done

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: I'm guessing ANF has the potential to interfere with beta reduction, and also with unfoldings.

For instance, `anf1 eta1` won't reduce, unless PLE substitutes `anf1` to get `(\x -> e) eta1`.

In a similar way, PLE could get stuck in `anf2 y` if PLE needs to substitute `anf2` to get a saturated application `reflectedFun x y`. I think this situation is not a problem when we don't want to reason with partial applications, but with partial applications, a term like
```Haskell
let anf2 = (\eta1 -> reflectedFun x eta1)
 in f anf2
```
could have `f` unfolded as follows
```Haskell
let anf2 = (\eta1 -> reflectedFun x eta1)
 in anf2 y
```
and now we would want `anf2` to be inlined so PLE learns that `eta1` is `y` which might allow additional unfoldings.

Some problems of the current implementation:
1. The above issues can happen with any let bindings in the code, not only ANF bindings.
2. `deANF` is applied over and over on multiple subexpressions and the bindings they use, redoing the inlining in the bindings, and potentially introducing more subexpressions, on which deANF would be applied again. It seems unnecessarily wasteful when the same work can be done by a single pass on the involved bindings. Granted, other issues might be uncovered when trying to deANF only once.
3. `deANF` seems to be inlining even when the bindings are not ANF bindings. Adding traces shows for instance that `VV##F##63` is rewritten to `SKIDC.CVar σ##a1wT γ##a1wU x##a1x2`. If only ANF bindings should be inlined, this would need to be fixed.

I don't have a proposal to fix (1) and (2), but I think it is at least worth documenting them in comments, probably users would care of them as well.

src/Language/Fixpoint/Types/Environments.hs
```diff
@@ -210,6 +210,9 @@ fromListIBindEnv = FB . S.fromList
 insertBindEnv :: Symbol -> SortedReft -> a -> BindEnv a -> (BindId, BindEnv a)
 insertBindEnv x r a (BE n m) = (n, BE (n + 1) (M.insert n (x, r, a) m))
 
+currentBindEnvId :: BindEnv a -> BindId
+currentBindEnvId (BE n _) = n
```
facundominguez: Maybe
```
currentBindEnvSize :: BindEnv a -> Int
currentBindEnvSize (BE n _) = n
```
?

AlecsFerra: Why? I would expect a BindEnv to be indexed by a BindId

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: 1 can be solved with local rewrites maybe
For 2 no idea

AlecsFerra: Also we could compile lets to lambdas

src/Language/Fixpoint/Types/Environments.hs
facundominguez: The commit that introduces the function says it returns the size of the BindEnv, and the implementation is returning the size of the BindEnv. In addition, it is unclear to me at least why the size needs to be framed as a BindId, and it is also unclear what bind such BindId would identify.

AlecsFerra: It's returning the last BindId inserted I need it for inserting the local rewrite

src/Language/Fixpoint/Solver/PLE.hs
facundominguez: I contributed a fix for (3) in https://github.com/ucsd-progsys/liquid-fixpoint/pull/711

AlecsFerra: SKIDC and the proposal are still passing? I would expect yes since it's a value

facundominguez: Your tests continue to pass with it.

facundominguez: I contributed another implementation that calls `deANF` only once per fork in the trie. It is passing all the tests as well. 2b8411e

In addition, it also explains the situations we have identified so far where ANF bindings interfere with PLE.

AlecsFerra: Yeah looks way better

src/Language/Fixpoint/Types/Environments.hs
facundominguez: Thanks, I'm starting to understand. Here are a few problems I see with such way of doing it.

I don't think that's how inserting binds works. The size is alway 1 unit higher than the last inserted bind.
```Haskell
insertBindEnv :: Symbol -> SortedReft -> a -> BindEnv a -> (BindId, BindEnv a)                                                                                                                
insertBindEnv x r a (BE n m) = (n, BE (n + 1) (M.insert n (x, r, a) m))
```

Also, [the definition](https://github.com/ucsd-progsys/liquid-fixpoint/blob/b4d217a80c58814db1a607237f73a0edf0f0d706/src/Language/Fixpoint/Types/Environments.hs#L101) of `BindEnv` says that all the BindIds inside the BindEnv should be smaller than the size.

Also, I don't think such expectation could hold after using `deleteBindEnv`.

And lastly, I think I would get immediately the purpose of the function if it was called `idOflastInsertedBind`.

My guess is that a comment would need to make clear that the binding might no longer be present in the environment if it was deleted before calling `idOflastInsertedBind`, and that it should return `the size - 1`. But isn't a less surprising API to just have a function that returns the size of the environment and call it `bindEnvSize`?

Update: Perhaps it also works to store somewhere the last id returned by `insertBindEnv`.

AlecsFerra: Ok fair tomorrow at the office I will implement your suggestion, I dont know about storing the index as the state is already pretty cluttered

facundominguez (approved):
LGTM! Any last changes before merging @AlecsFerra?

AlecsFerra:
I think we can merge it

facundominguez:
> 1. The above issues can happen with any let bindings in the code, not only ANF bindings.

From the open discussion above, this is the only standing question, but I think it would be fine to handle it in some other PR.

PR #712, opened 2024-10-16
Only produce `.smt2` files when the `save` flag is turned on

gergoerdi:
(See https://github.com/ucsd-progsys/liquidhaskell/issues/2357)

### DIFF
--- a/src/Language/Fixpoint/Smt/Interface.hs
+++ b/src/Language/Fixpoint/Smt/Interface.hs
@@ -60,11 +60,7 @@ module Language.Fixpoint.Smt.Interface (
     ) where
 
 import           Language.Fixpoint.Types.Config ( SMTSolver (..)
-                                                , Config
-                                                , solver
-                                                , smtTimeout
-                                                , gradual
-                                                , stringTheory)
+                                                , Config (solver, smtTimeout, gradual, stringTheory, save))
 import qualified Language.Fixpoint.Misc          as Misc
 import           Language.Fixpoint.Types.Errors
 import           Language.Fixpoint.Utils.Files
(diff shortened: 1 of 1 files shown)


src/Language/Fixpoint/Smt/Interface.hs
```diff
@@ -240,12 +236,17 @@ negativeP
...
+           hLog <- openFile smtFile WriteMode
+           hSetBuffering hLog $ BlockBuffering $ Just $ 1024 * 1024 * 64
+           return $ Just hLog
+       me   <- makeContext' cfg mb_hLog
        pre  <- smtPreamble cfg (solver cfg) me
-       mapM_ (\l -> SMTLIB.Backends.command_ (ctxSolver me) l >> BS.hPutBuilder hLog l >> LBS.hPutStr hLog "\n") pre
+       mapM_ (SMTLIB.Backends.command_ (ctxSolver me)) pre
+       forM_ mb_hLog $ \hLog -> forM_ pre $ \l -> do
+           BS.hPutBuilder hLog l
+           LBS.hPutStr hLog "\n"
```
facundominguez: If there is no particular reason for the current arrangement, maybe it is worth preserving the order in which commands are sent one at a time to both the SMT solver and the log, in case the SMT solver crashes.

facundominguez (commented):
It is looking good to me. I shared just one comment about the order in which things are written to the log.

src/Language/Fixpoint/Smt/Interface.hs
gergoerdi: Yeah, I was debating with myself on whether I should be conservative and do that, or write somewhat simpler code. So now its 2:1 in favor of being conservative :), hence I updated the PR.

facundominguez (approved):
LGTM!

PR #715, opened 2024-10-24
Remove the preference for ∧ as the dedicated monoid on Preds

shingarov:
While the Preds certainly do form a monoid under ∧, I don't think this monoid is "better" then other monoids of Preds (e.g. the one under ∨), so there is no compelling reason to dedicate ∧ as *the* implementation of <> in Pred.  It is also not used anywhere in the codebase.

### DIFF
--- a/src/Language/Fixpoint/Horn/Types.hs
+++ b/src/Language/Fixpoint/Horn/Types.hs
@@ -73,12 +73,6 @@ data Pred
   deriving (Data, Typeable, Generic, Eq, ToJSON, FromJSON)
 
 
-instance Semigroup Pred where
-  p1 <> p2 = PAnd [p1, p2]
-
-instance Monoid Pred where
-  mempty = Reft mempty
-
 instance F.Subable Pred where
   syms (Reft e)   = F.syms e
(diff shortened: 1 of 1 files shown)


facundominguez (approved):
Thanks @shingarov. Looks good to me. We will have to check if the Monoid instance is used in LH.

shingarov:
> check if … used in LH

Thanks!  While we are at it, can I ask you a totally-silly question?  What is the best-practice workflow for trying out effects of LF changes on LH?  Like, if I make some change in LF and I want to see what happens to, say, `liquidhaskell/tests/benchmarks/icfp15/MiniPosix.hs`, which way to rebuild would be your "preferred" way?

facundominguez:
What I would try is
```bash
cabal build liquidhaskell
cabal exec ghc -- -fplugin=LiquidHaskell -fforce-recomp liquidhaskell/tests/benchmarks/icfp15/MiniPosix.hs
```
after modifying `liquid-fixpoint`.

facundominguez:
LH doesn't seem to depend on the removed instance.

PR #716, opened 2024-10-31
Unfold with undecided guards in PLE when new equalities are discovered

facundominguez:


facundominguez:
Merged this too soon. `tests/ple/pos/Permutations.hs` is the only test dramatically slowing down from 6 seconds to 18 seconds. I'm benchmarking an alternative fix.

PR #718, opened 2024-10-31
Unfold with undecided guards in PLE (take 2)

facundominguez:
Follow up to #716. Instead of collecting new equalities, which might be already known, test for equality the result of evaluating the guards. This measure is more conservative with the unfoldings, still makes the tests pass, and doesn't cause a performance regression for tests/ple/pos/Permutations.hs.


facundominguez:
top
bot

PR #719, opened 2024-11-08
Remove some more monoid instances

facundominguez:
In the same direction as #715, this PR removes monoids for `Reft`, `Expr`, `Sort`, and `SortedReft`.


facundominguez:
I checked that LH can be easily updated to build without these.

PR #723, opened 2024-11-30
Parameterize EquationVs eqName field type

facundominguez:
This makes possible to rename the name of the equation with a transformation that changes its type.


facundominguez:
I reverted this in 2acb214 soon after merging. I didn't need it in the end, and I don't have other measures by which the representation can be considered better.

PR #724, opened 2024-12-02
Fix sorts of bag theory operations in Theories.hs

facundominguez:


facundominguez:
@nikivazou, @clayrat, could we have some tests for these sorts?

I only discovered they were wrong when I tried to use them in LH to build the environment to resolve their names.

PR #725, opened 2024-12-12
Support for lambdas :)

AlecsFerra:


AlecsFerra:
This should be pretty definitive even if it's useless without the corresponding PR on LH

src/Language/Fixpoint/Solver/PLE.hs
```diff
@@ -653,6 +653,26 @@ eval γ ctx et = go
 -- | 'evalELamb' produces equations that preserve the context of a rewrite
 -- so equations include any necessary lambda bindings.
 evalELam :: Knowledge -> ICtx -> EvalType -> (Symbol, Sort) -> Expr -> EvalST (Expr, FinalExpand)
+evalELam γ ctx et (x, s) e
+  | not $ isEtaSymbol x = do
+    -- We need to refresh it as for some reason lamdba variables
+    -- as reflected lambdas variables for some reason are declared
+    -- two times, maybe we should define a new type of identifier
+    -- and not reuse the etabeta ones
```
```suggestion
    -- We need to refresh it as for some reason lamdba variables,
    -- as reflected, are declared twice. Maybe we should define
    -- a new type of identifier and not reuse the etabeta ones
```

facundominguez: Can we have an example of what it means for a lambda variable to be reflected twice?

It looks like it would be worth investigating that, instead of changing the code here to workaround a possible bug.

src/Language/Fixpoint/Smt/Theories.hs
```diff
@@ -590,7 +590,7 @@ interpSym :: Symbol -> Raw -> Sort -> (Symbol, TheorySymbol)
 interpSym x n t = (x, Thy x n t Theory)
 
 maxLamArg :: Int
-maxLamArg = 7
+maxLamArg = 20
```
facundominguez, line 599: This `maxLamArg` looks rather pointless, it is not used for anything other than exporting it. If you know what is needed for, it would be helpful to document it here, as well as what is the consequence of using lambdas that exceed this amount of arguments.

facundominguez (commented):
Thanks @AlecsFerra. A spot that could need more discussion are the changes to `evalELam`.

src/Language/Fixpoint/Smt/Theories.hs
AlecsFerra: I added a comment to explain how it's used or at least what I understood about it, I don't know if it's used somewhere else

src/Language/Fixpoint/Solver/PLE.hs
AlecsFerra: Idk why but if for example you take the SKILam and remove this branch it will complain that the variable name that is used by the lambdas is declared twice in the smt file

facundominguez: Tried to reproduce this in a smaller example, but couldn't. But I think it is ok to investigate it separately.

facundominguez (approved):

PR #729, opened 2024-12-18
Clean up support for CVC5

### LINKED ISSUE #728: fixpoint crashes when using CVC
Fixpoint crashes when running on the following with cvc4 and cvc5

[calc_cample_point_err.smt2](https://github.com/user-attachments/files/18184905/calc_cample_point_err.txt)

with `fixpoint --solver=cvc4`:
```
Crash!: :1:1-1:1: Error
  crash: SMTLIB2 respSat = Error "Parse Error: Symbol 'Z3_OP_MUL' not declared as a variable"
CRASH: -1 : ""
```

with `fixpoint --solver=cvc5`:

```
fixpoint: fd:14: hPut: resource vanished ([smtlib-backends-process] while sending command "<long string containing the smtlib2 file>" to the solver: Broken pipe
```

The content of the string is:

[error.smt2](https://github.com/user-attachments/files/18184926/error.txt)

Running cvc5 directly on that also results in a error related to `Z3_OP_MUL`

ranjitjhala:
- Simplifies the "preamble" code
- Fixes some glitches in serialization
- Adds `cvc5` to the CI ... (!!!)

Fixes #728

### DIFF
--- a/src/Language/Fixpoint/Smt/Theories.hs
+++ b/src/Language/Fixpoint/Smt/Theories.hs
@@ -375,14 +375,14 @@ sortAppInfo t = case bkFFunc t of
 
 -- | `theorySymbols` contains the list of ALL SMT symbols with interpretations,
 --   i.e. which are given via `define-fun` (as opposed to `declare-fun`)
-theorySymbols :: Config -> [DataDecl] -> SEnv TheorySymbol -- M.HashMap Symbol TheorySymbol
+theorySymbols :: SMTSolver -> [DataDecl] -> SEnv TheorySymbol -- M.HashMap Symbol TheorySymbol
 theorySymbols cfg ds = fromListSEnv $  -- SHIFTLAM uninterpSymbols
                                   interpSymbols cfg
                                ++ concatMap dataDeclSymbols ds
 
 
 --------------------------------------------------------------------------------
-interpSymbols :: Config -> [(Symbol, TheorySymbol)]
+interpSymbols :: SMTSolver -> [(Symbol, TheorySymbol)]
 --------------------------------------------------------------------------------
 interpSymbols cfg =
   [
@@ -526,11 +526,10 @@ interpSymbols cfg =
                                          (bagSort $ FVar 0)
     bagSubSort = FAbs 0 $ FFunc (bagSort $ FVar 0) $ FFunc (bagSort $ FVar 0) boolSort
 
-bv2i :: Config -> Int -> Raw
-bv2i cfg size = case solver cfg of
-  Cvc4 -> "bv2nat"
-  Cvc5 -> "bv2nat"
-  _    -> Data.Text.pack $ printf "(_ bv2int %d)" size
+bv2i :: SMTSolver -> Int -> Raw
+bv2i Cvc4 _ = "bv2nat"
+bv2i Cvc5 _ = "bv2nat"
+bv2i _    n = Data.Text.pack $ printf "(_ bv2int %d)" n
 
 interpBvUop :: Symbol -> (Symbol, TheorySymbol)
 interpBvUop name = interpSym' name bvUopSort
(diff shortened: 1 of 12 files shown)


src/Language/Fixpoint/Smt/Theories.hs
```diff
@@ -515,6 +525,13 @@ interpSymbols =
                                  $ FFunc (bagSort $ FVar 0)
                                          (bagSort $ FVar 0)
     bagSubSort = FAbs 0 $ FFunc (bagSort $ FVar 0) $ FFunc (bagSort $ FVar 0) boolSort
+
+bv2i :: Config -> Int -> Raw
+bv2i cfg size = case solver cfg of
+  Cvc4 -> "bv2nat"
+  Cvc5 -> "bv2nat"
```
nilehmann: Is this correct? Sounds like `bv2nat` has different semantics than `bv2int`

ranjitjhala: https://github.com/Z3Prover/z3/issues/1252

ranjitjhala: apparently they do the same thing... (per Nikolaj B.)

nilehmann: `cvc` people say is not standard smtlib https://github.com/cvc5/cvc5/issues/11252, but I see it here https://smt-lib.org/theories-FixedSizeBitVectors.shtml

ranjitjhala: Yes, its odd. Furthermore -- CVC seems to just _not support_ `bv2int` -- I get `Symbol 'bv2int' not declared as a variable` errors...? I guess you better only implicitly convert `nat`s into `bv`...

ranjitjhala: Btw, https://github.com/Z3Prover/z3/pull/6419 seems to indicate that even in z3 `bv2int` and `bv2nat` do the same thing?

nilehmann: Oh I get it now, `bv2int` and `bv2nat` are not part of the standard they are only defined to define the semantics. From the CVC5 issue it seems that is going to be standardized in smtlib 2.7 so maybe worth tracking in an issue. 

Also is bv2int/bv2nat in z3 the same as bv2nat in CVC?

nilehmann: Ok I get it now (for real)

* In Z3, `bv2int` and `bv2nat` are aliases. They both implement the semantics of the upcoming `bv2nat`, i.e., they interpret the bit vector as an unsigned integer
* In CVC, only `bv2nat` is implemented and it has the semantics of the upcoming `bv2nat`

In smtlib 2.7, `bv2int` will have a precise meaning (bit vectors as signed integer) different from the one implemented by z3, so the correct thing here is to use `bv2nat` both for z3 and cvc because it will be forward compatible.

Fixpoint (and users of fixpoint) should probably start using `*_to_nat` instead of `*_to_int`  to reserve `to_int` for the upcoming meaning in 2.7.

src/Language/Fixpoint/Smt/Theories.hs
```diff
@@ -477,14 +486,15 @@ interpSymbols =
...
   , interpBvCmp bvSGtName
   , interpBvCmp bvSGeName
-
-  , interpSym intbv32Name "(_ int2bv 32)"   (FFunc intSort bv32)
-  , interpSym intbv64Name "(_ int2bv 64)"   (FFunc intSort bv64)
-  , interpSym bv32intName  "(_ bv2int 32)"  (FFunc bv32    intSort)
-  , interpSym bv64intName   "(_ bv2int 64)" (FFunc bv64    intSort)
-
+  , interpSym intbv32Name   "(_ int2bv 32)" (FFunc intSort bv32)
+  , interpSym intbv64Name   "(_ int2bv 64)" (FFunc intSort bv64)
```
nilehmann, line 490: probably a similar argument applies here and we should use `nat2bv`, but I haven't checked.

PR #730, opened 2025-01-13
Instantiate negative tyvars

AlecsFerra:
I don't have a small file to reproduce the issue but was triggered by @pacastega 's project

### DIFF
--- a/src/Language/Fixpoint/SortCheck.hs
+++ b/src/Language/Fixpoint/SortCheck.hs
@@ -72,6 +72,7 @@ import           Control.Monad.Reader
 
 import           Data.Bifunctor (first)
 import qualified Data.HashMap.Strict       as M
+import qualified Data.HashSet              as S
 import           Data.IORef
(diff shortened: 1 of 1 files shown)


src/Language/Fixpoint/SortCheck.hs
```diff
@@ -960,9 +961,26 @@ throwErrorAt ~err' = do -- Lazy pattern needed because we use LANGUAGE Strict in
 -- | Helper for checking symbol occurrences
 checkSym :: Env -> Symbol -> CheckM Sort
 checkSym f x = case f x of
-  Found s -> instantiate s
+  Found s -> refreshNegativeTyVars s >>= instantiate
   Alts xs -> throwErrorAt (errUnboundAlts x xs)
 
+-- Negative type variables are implictly universally quantified type variables
```
facundominguez, line 967: Insight like this should be explained in the documentation of `Sort` data type, so people know how to use it to represent sorts.

The code itself looks good.

PR #731, opened 2025-01-16
Delay applyExpr until the very end of elaboration

vrindisbacher:


vrindisbacher:
hmm @ranjitjhala looks like the auto formatting went crazy and reformatted almost all of `SortCheck.hs. Is there a specific formatter you're using?

PR #732, opened 2025-01-17
Delay applyExpr until after elaboration

vrindisbacher:
There are a few changes I'm hoping to implement here. There are: 

1. delaying use of `applyExpr` in `elab` until after all the elaboration has taken place. 
2. delaying use of `apply` in `elab` until after all the elaboration has taken place - hoping to use union find to do this. See [the PoC implementation here](https://github.com/vrindisbacher/toy-union-find/blob/main/app/Elab.hs). 
3. Separate type visitor and type folder into two separate structures. Additionally, ditch using the `State` monad for folding and replace it with the `Reader` monad - similar to what is implemented for `CheckM`

The reasons for these changes are performance hang ups when passing large queries to fixpoint. We noticed that a significant amount of time is being spent in `elab` - and we suspect it's due to the exponential nature of using `apply` and `applyExpr`. Additionally, we observed that `trans` and `fold` in the type visitors and folders were taking up a massive amount of memory. By "specializing" the visitor not to have an accumulator we reduced the memory footprint on certain queries by ~ half. 

As of now, changes 1 and 3 are implemented and I'm working on 2. Just 1 and 3 cut down the time fixpoint takes on large queries by ~ 1/3.

### DIFF
--- a/src/Language/Fixpoint/Types/Visitor.hs
+++ b/src/Language/Fixpoint/Types/Visitor.hs
@@ -57,6 +57,9 @@ import qualified Data.HashMap.Strict as M
 import qualified Data.List           as L
 import           Language.Fixpoint.Types hiding (mapSort)
 import qualified Language.Fixpoint.Misc as Misc
+import Control.Monad.Reader
+import GHC.IO (unsafePerformIO)
+import Data.IORef (newIORef, readIORef, IORef, modifyIORef')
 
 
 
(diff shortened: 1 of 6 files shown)


ranjitjhala:
let me check this works with the LH tests too...

vrindisbacher:
Sounds good - fwiw I also ran the flux tests with this version and everything passed.

nilehmann:
I'm trying to rationalize the correctness of this change. 

I think you can delay the substitution of sort variables as long as you don't structurally match on sorts, i.e., you don't make decisions based on the structure of sorts (which may have solved variables that should be considered in the decision). Presumably, elaboration shouldn't be making decisions based on the structure of sorts in the expression being elaborated so this change sounds fine...

...except that you structurally match on sorts when printing expressions for errors. For example, I think this changes the behavior of this line https://github.com/ucsd-progsys/liquid-fixpoint/blob/ac79d5ca46c02b6861d54537f0f0cc2762edce92/src/Language/Fixpoint/SortCheck.hs#L586 because `checkIteTy` prints the expressions (with pending substitutions)  in case of an error. However, I have the impression that this always has been wrong and errors may print expressions with sorts variables that have already been solved.

unrelated and out of curiosity: do the changes in strictness have any impact?  the constructors of `Expr` are already marked as strict

vrindisbacher:
Ah ya you're right - I suppose you could use `applyExpr` in the error case of `checkIteTy`. e..g `errIte e1 e2 t1 t2` -> `errIte (applyExp theta e1) (applyExp theta e2) t1 t2`?

vrindisbacher:
I think there are some bugs in `Foldable`. Working them out now.

src/Language/Fixpoint/Types/Visitor.hs
```diff
@@ -270,13 +372,12 @@ mapMExpr f = go
     go (PAnd ps)       = f . PAnd =<< (go `traverse` ps)
     go (POr ps)        = f . POr =<< (go `traverse` ps)
 
-mapKVarSubsts :: Visitable t => (KVar -> Subst -> Subst) -> t -> t
-mapKVarSubsts f          = trans kvVis () ()
+mapKVarSubsts :: Foldable t => (KVar -> Subst -> Subst) -> t -> t
+mapKVarSubsts f          = trans txK
```
vrindisbacher, line 430: There seems to be a bug here but I'm not sure what it could be. 

Also, the tests only fail if you set `sat.smt = false` with `z3` but the `smt2` file being thrown at fixpoint is different so there is definitely a problem with (I'm assuming) `Foldable`. An example test that fails is  `tests/pos/test2.fq`

vrindisbacher: Fixed!

facundominguez:
Hello! Is there a description available of the problem that this contribution is addressing? I don't promise to review, but it would be helpful for future reference nonetheless.

vrindisbacher:
> Hello! Is there a description available of the problem that this contribution is addressing? I don't promise to review, but it would be helpful for future reference nonetheless.

I just updated the description with some details. Let me know if you have questions / if I can provide additional details.

PR #734, opened 2025-01-22
conditional elaboration of theories for cvc5

clayrat:


src/Language/Fixpoint/Smt/Theories.hs
```diff
@@ -519,7 +530,7 @@ interpSymbols cfg =
                                  $ FFunc (FVar 1)
                                          (mapSort (FVar 0) (FVar 1))
 
-    bagCountSort = FAbs 0 $ FFunc (bagSort $ FVar 0) $ FFunc (FVar 0) intSort
+    bagCountSort = FAbs 0 $ FFunc (FVar 0) $ FFunc (bagSort $ FVar 0) intSort
```
facundominguez: This is likely to cause wreckage in LH. The arguments needed to be already flipped in 9172c8fad for #724.

At this point, it seems we need to test this for both cvc5 and z3.

clayrat: Yes, this is a breaking change, I will update the corresponding `assume`/`define` in LH when I'm done with this PR. On the flip side, the `Set_add` is as of now unused in LH, I'll add it as well.

The reason these have to flipped is that we don't do any elaboration at all for CVC5, mapping LF theories onto it directly. The alternative is to introduce a micro-elaboration step for CVC5 just to flip the arguments for `Bag_count`.

PR #735, opened 2025-02-05
Elab Optimizations

vrindisbacher:
A few changes are implemented here: 

- delaying use of `applyExpr` in elab until after all the elaboration has taken place.
- Separate type visitor and type folder into two separate structures. Additionally, ditch using the State monad for folding and replace it with the Reader monad - similar to what is implemented for `CheckM`
- Ripping out a redundant `unifyExpr`. This seemed to be re-traversing the expression tree and creating type substitutions by unifying sorts that were already unified in `elabEAppSort`.


ranjitjhala:
sweet -- let me try this out on LH to check for (performance) regressions. Btw, @vrindisbacher you ran the `flux` tests yes?

ranjitjhala:
@vrindisbacher (I gave you push access to `liquid-fixpoint`) 

is there some way to modify this PR so instead it is pulled into a `elab-opt` branch in the main `ucsd-progsys/liquid-fixpoint`? 

then I make a branch on LH to test on the CI etc.

vrindisbacher:
Yea let me see. Shouldn't be too hard

vrindisbacher:
@ranjitjhala see #736. I'm going to close this PR.

PR #736, opened 2025-02-05
Optimize Elaboration 

vrindisbacher:
@ranjitjhala, this one is from the liquid-fixpoint origin but has the same changes as #735 

(copying over notes)

A few changes are implemented here:

* Delaying use of `applyExpr` in `elab` until after all the elaboration has taken place.
* Separate type `Visitor` and type `Folder` into two separate structures. 
* Additionally, ditch using the `State` monad for folding and replace it with the `Reader` monad - similar to what is implemented for `CheckM`
* Ripping out a redundant `unifyExpr`. This seemed to be re-traversing the expression tree and creating type substitutions by unifying sorts that were already unified in `elabEAppSort`.

[RJ: All of these together yield _massive_ speedups for some `flux` benchmarks]

### DIFF
--- a/liquid-fixpoint.cabal
+++ b/liquid-fixpoint.cabal
@@ -176,6 +176,8 @@ library
     ghc-options: -Wno-unused-imports
   if impl(ghc>9.8)
     ghc-options: -Wno-x-partial
+  if impl(ghc>9.10)
+    ghc-options: -Wno-deriving-typeable
   if flag(devel)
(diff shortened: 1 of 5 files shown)


ranjitjhala:
Thanks @vrindisbacher -- seems to give some nice improvements in the LH benchmarks too!

vrindisbacher:
Awesome!

PR #740, opened 2025-03-08
add bitwuzla as a solver

vrindisbacher:
Adding bitwuzla as a solver because it's supposedly good at bitvectory stuff!

### DIFF
--- a/src/Language/Fixpoint/Smt/Interface.hs
+++ b/src/Language/Fixpoint/Smt/Interface.hs
@@ -307,6 +307,10 @@ makeContext' cfg ctxLog
                       Process.defaultConfig
                              { Process.exe = "cvc5"
                              , Process.args = ["--incremental", "-L", "smtlib2"] }
+         Bitwuzla    -> makeProcess ctxLog $
+                      Process.defaultConfig
+                             { Process.exe = "bitwuzla"
+                             , Process.args = ["--lang", "smt2"] }
        solver <- SMTLIB.Backends.initSolver SMTLIB.Backends.Queuing backend
        loud <- isLoud
        return Ctx { ctxSolver    = solver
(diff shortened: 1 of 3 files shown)


ranjitjhala:
Any idea why the tests are failing? looks like some odd Circle/CI hassle? 

The new solver doesn't look like it should affect the existing tests at all?

src/Language/Fixpoint/Smt/Interface.hs
```diff
@@ -459,7 +463,8 @@ makeMbqi cfg
 z3_options :: [Builder]
 z3_options
   = [ "(set-option :auto-config false)"
-    , "(set-option :model true)" ]
+    , "(set-option :model true)"
+    , "(set-option :sat.smt true)" ]
```
nilehmann: I thought we were told not to use this.

nilehmann: circle ci is using an ancient version of z3, 4.8.7, released in 2019. That's probably why adding this line causes it to fail. For the gh one, you need to update the cache action. Also, why do you have both circle ci and gh actions running?

vrindisbacher: Oh whoops - I did not mean to push that. That is stuff I usually add locally so that `tock-veri-asm` checks faster. I'll patch it now.

nilehmann: The GitHub z3's version is also a tiny bit old (4.10.2 released in 2022)

ranjitjhala:
@vrindisbacher -- did this bitwuzla thing help? should we try to merge this?

vrindisbacher:
Honestly I haven't had time to take a look. Maybe something to close for now and we can come back to it?

PR #748, opened 2025-03-28
Implement let bindings

nilehmann:
We've found significant performance improvements when reducing the size of expressions sent to fixpoint (and the smt solver) by deduplicating common subexpressions using let bindings. For simplicity, this only implements one binder per let.

The first commit implements the parts I'm more confident about. The second commit touches PLE stuff that I barely understand, so please review that more carefully.

There are also a couple of functions that match on an expression that I didn't modify, so they don't show up in the PR:

* Language.Fixpoint.Solver.Instantiate.topApps 
* Language.Fixpoint.Solver.PLE.isExprRewritable
* Language.Fixpoint.Solver.PLE.simplify
* Language.Fixpoint.Solver.Rewrite.convert

### DIFF
--- a/src/Language/Fixpoint/Defunctionalize.hs
+++ b/src/Language/Fixpoint/Defunctionalize.hs
@@ -85,6 +85,9 @@ normalizeLamsFromTo i   = go
                               (i2, e2') = go e2
                           in (max i1 i2, EApp e1' e2')
     go (ECst e s)       = fmap (`ECst` s) (go e)
+    go (ELet x e1 e2)   = let (i1, e1') = go e1
+                              (i2, e2') = go e2
+                          in (max i1 i2, ELet x e1' e2')
     go (EIte e1 e2 e3)  = let (i1, e1') = go e1
                               (i2, e2') = go e2
                               (i3, e3') = go e3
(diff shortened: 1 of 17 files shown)


facundominguez:
At the moment I see the following challenges for PLE,

1. for rest-rewrite, when we have `let y = (g x) in f y`, we still want to spot the opportunity to apply a rule to rewrite `f . g` if such rule exists.
2. when deciding whether to unfold `let y = e in f y`, `y = e` needs to be provided to the SMT solver in the queries that test the guards of `f`.
3. when PLE decides to unfold `f` in `let y = e in f y`, it produces an equality `f y = body_of_f_y` which will require the accompanying equality `y = e`.

(1) could be achieved by inlining the let bindings in the expressions given to rest-rewrite. (2) and (3) could be achieved by collecting the local bindings in an environment to use at the time of the queries and when producing equalities for unfoldings (or to be designed, maybe sending them to the SMT solver as they are found). It could make things simpler if we can be sure that no two let bindings have the same name.

ranjitjhala:
Excellent — thanks for catching those subtle points @facundominguez ! However I think this PR shouldn’t break anything currently (since the PLE functions didn’t use let — those were already inlined IIRC?) 

i will check with the LH tests…

ranjitjhala (approved):
Looks great, I'll try out with LH and then merge!

ranjitjhala:
passes LH with some minor edits!

PR #752, opened 2025-05-13
generate apply tags dynamically

clayrat:
This PR removes `Language.Fixpoint.Types.Theories.funcSorts`, which was an ad-hoc function pre-generating all possible sort combinations for applys statically. Instead, new apply definitions are now accumulated in a state monad and emitted dynamically next to the declarations that require them, as late as possible. This is achieved by tracking the definitions in a stack of sort maps, where the nesting level corresponds to the SMT push/pop level.

All SMT symbol generation is now done in a `SymM` monad responsible for tracking the definition stack, which lifts into `SmtM` monad, used for interacting with the SMT solver. The `SmtM` monad itself is lifted into `SolveM` and `EvalST` monads where necessary.

Also, a heuristic has been added that suppresses the generation of `lam_arg` definitions unless the `allowHO` flag is set. This is sufficient to pass all tests in LF and LH, but adjustments to the heuristic may be necessary for larger codebases.

### DIFF
--- a/src/Language/Fixpoint/Smt/Types.hs
+++ b/src/Language/Fixpoint/Smt/Types.hs
@@ -12,8 +12,11 @@ module Language.Fixpoint.Smt.Types (
     -- * Serialized Representation
     --    symbolBuilder
 
+    -- * SMT monad
+      SmtST
+
     -- * Commands
-      Command  (..)
+    , Command  (..)
 
     -- * Responses
     , Response (..)
@@ -27,6 +30,7 @@ module Language.Fixpoint.Smt.Types (
 
     ) where
 
+import           Control.Monad.State
 import           Data.ByteString.Builder (Builder)
 import           Language.Fixpoint.Types
 import           Language.Fixpoint.Types.Config (ElabFlags)
@@ -43,6 +47,9 @@ import           System.IO                (Handle)
 
 -- symbolBuilder :: Symbol -> LT.Builder
 -- symbolBuilder = LT.fromText . symbolSafeText
+
+-- | SMT monad
+type SmtST a = StateT SymEnv IO a
 
 -- | Commands issued to SMT engine
 data Command      = Push
@@ -110,7 +117,7 @@ data Context = Ctx
 --------------------------------------------------------------------------------
 
 class SMTLIB2 a where
-  smt2 :: SymEnv -> a -> Builder
+  smt2 :: a -> State SymEnv Builder
 
-runSmt2 :: (SMTLIB2 a) => SymEnv -> a -> Builder
+runSmt2 :: (SMTLIB2 a) => a -> State SymEnv Builder
 runSmt2 = smt2
--- a/src/Language/Fixpoint/Smt/Types.hs
+++ b/src/Language/Fixpoint/Smt/Types.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleInstances         #-}
+{-# LANGUAGE TupleSections             #-}
 {-# LANGUAGE FlexibleContexts          #-}
 {-# LANGUAGE NoMonomorphismRestriction #-}
 {-# LANGUAGE OverloadedStrings         #-}
@@ -12,11 +13,8 @@ module Language.Fixpoint.Smt.Types (
     -- * Serialized Representation
     --    symbolBuilder
 
-    -- * SMT monad
-      SmtST
-
     -- * Commands
-    , Command  (..)
+      Command  (..)
 
     -- * Responses
     , Response (..)
@@ -28,9 +26,16 @@ module Language.Fixpoint.Smt.Types (
     -- * SMTLIB2 Process Context
     , Context (..)
 
-    ) where
+    -- * SMT monad
+    , SmtM
+    , hoistSMT
+    , catchSMT
+    , bracketSMT
 
-import           Control.Monad.State
+    ) where
+import           Control.Exception
+-- import           Control.Monad.State
+import           Control.Monad.Reader
 import           Data.ByteString.Builder (Builder)
 import           Language.Fixpoint.Types
 import           Language.Fixpoint.Types.Config (ElabFlags)
@@ -47,9 +52,6 @@ import           System.IO                (Handle)
 
 -- symbolBuilder :: Symbol -> LT.Builder
 -- symbolBuilder = LT.fromText . symbolSafeText
-
--- | SMT monad
-type SmtST a = StateT SymEnv IO a
 
 -- | Commands issued to SMT engine
 data Command      = Push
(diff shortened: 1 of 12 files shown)


clayrat:
Not sure why CircleCI is failing.

facundominguez:
Circleci uses the same stackage snapshot as the job `stack/ghc-9.10.1` and different versions of system dependencies (e.g. z3). But it might be more likely that the build fails because of differences in PLE implementations. The circle CI job runs tests with --interpreter when it fails.

ranjitjhala:
Many thanks @clayrat -- this was long overdue!!! (Sorry I was traveling last week, please let me know if there is something I should take a look at!)

clayrat:
> Circleci uses the same stackage snapshot as the job `stack/ghc-9.10.1` and different versions of system dependencies (e.g. z3). But it might be more likely that the build fails because of differences in PLE implementations. The circle CI job runs tests with --interpreter when it fails.

Oh, I guess some declarations are missing in `Language.Fixpoint.Solver.Interpreter`, let me try debugging the `--interpreter` mode.

clayrat:
> Many thanks @clayrat -- this was long overdue!!! (Sorry I was traveling last week, please let me know if there is something I should take a look at!)

I'm not sure if there's something in particular, just wondering whether you like the proposed solution in general. :)

clayrat:
All green now.

src/Language/Fixpoint/Smt/Types.hs
```diff
@@ -102,15 +110,38 @@ data Context = Ctx
   , ctxLog     :: !(Maybe Handle)
   , ctxVerbose :: !Bool
   , ctxSymEnv  :: !SymEnv
+  , ctxIxs     :: ![Int]
   , ctxDefines :: DefinedFuns
+  , ctxLams    :: !Bool
```
facundominguez, line 118: Could we have documentation for these new fields?

src/Language/Fixpoint/Solver/Instantiate.hs
```diff
@@ -801,13 +808,13 @@ assertSelectors γ expr' = do
...
-withCtx cfg file env defns k = do
-  ctx <- SMT.makeContextWithSEnv cfg file env defns
-  _   <- SMT.smtPush ctx
-  res <- k ctx
-  SMT.cleanupContext ctx
-  return res
+withCtx :: Config -> FilePath -> SymEnv -> DefinedFuns -> SmtM a -> IO a
+withCtx cfg file env defns k =
+  bracket acquire release $
+    evalStateT $ SMT.smtPush >> k   -- TODO why is there no pop?
```
facundominguez, line 814: I'd rather ask why is there a push? There might be some calling convention at work, most discoverable if we actually remove the push and see what happens.

The pop is probably unnecessary if the context is destroyed immediately.

src/Language/Fixpoint/Types/Theories.hs
```diff
@@ -70,41 +78,65 @@ type Raw = Text
...
+  , seAppls    :: !Appls                    -- ^ Stack of function sort maps
+  , seApplsCur :: !(M.HashMap FuncSort Int) -- ^ Current function sort map
+  , seIx       :: !Int                      -- ^ Largest unused index for sorts
   }
   deriving (Eq, Show, Data, Typeable, Generic)
 
 {- type FuncSort = {v:Sort | isFFunc v} @-}
 type FuncSort = (SmtSort, SmtSort)
 
+type SymM a = State SymEnv a
```
facundominguez, line 133: Would be useful to have some comments explaining the purpose of SymM.

facundominguez (commented):
Thanks @clayrat. I added some suggestions of things to explain. Regarding the current solution, I think that a blob of documentation explaining the API to generate applies would be helpful to future maintainers.

Questions that I have in mind:
* What is an apply?
* How to register the need to generate an apply?
* How to generate an apply that is necessary?
* When to call `clearApplies`?
* What is lam_arg?
* What is the relation between lam_arg and applies that causes lam_arg to be disabled without --higherorder?

Showing short examples will probably help.

Also, I'm guessing one change in this PR must be that applies were pregenerated once in a preamble, whereas now they might be produced more than once at the places where they are needed.

I don't see anything amiss in the code itself, except that I need some more context to follow the details. :)

src/Language/Fixpoint/Solver/Solve.hs
```diff
@@ -132,7 +134,8 @@ solve_ cfg fi s0 ks wkl = do
   (fi1, s4, res1) <- case resStatus res0 of  {- first run the interpreter -}
     Unsafe _ bads | not (noLazyPLE cfg) && rewriteAxioms cfg && interpreter cfg -> do
       fi1 <- doInterpret cfg fi (map fst $ mytrace ("before the Interpreter " ++ show (length bads) ++ " constraints remain") bads)
-      (s4, res1) <-  sendConcreteBindingsToSMT F.emptyIBindEnv $ \bindingsInSmt -> do
+      clearApplys
```
facundominguez, line 142: This is some new action at a distance, and needs to make clients of the SMT solver aware of an implementation detail. Perhaps it is unavoidable, but I couldn't tell before understanding the problem and the solution a bit better.

clayrat:
@facundominguez I've added some comments to the code, but I'm not sure where to include the answers to your questions. As a first step, maybe I can post them here and we can decide where to add the necessary parts.

> * What is an apply?
> * What is lam_arg?

The concepts of 'apply', 'lam_arg', 'coerce' and 'lambda' are not new to the PR; what has changed is how they are generated. The high-level idea is that, during solving and elaboration, higher-order functions are defunctionalised into various combinations of special symbols (`apply` and the other three), which are then dispatched according to the type they act on. We then further serialise these function types into tags and define the defunctionalization symbols (e.g. `(declare-fun apply$35$$35$1 (Int Bool) Int)` - here `1` corresponds to `Bool -> Int`, and the first `Int` of the tuple is the number of the function the `apply` is acting on). The entry in the resulting lookup table that matches a function type (e.g. `Bool -> Int`) with a serialised tag (e.g. `1`) is what I colloquially (and perhaps confusingly) refer to as an 'apply'. Before this PR, such a tag lookup table was pre-generated in the preamble; however, there were at least a couple of issues with that approach. 
1. PLE would sometimes use types from imported measures to generate additional tags in the '.evals' file, causing an 'unknown function/constant' crash when returning to the main solver routine (this occurred in `pos/Maps` and `ple/pos/PolySetBar`, as mentioned in https://github.com/ucsd-progsys/liquidhaskell/issues/2438). 
2. Due to recent improvements in polymorphism handling, the static tag table increased in size considerably when creating all possible combinations of polymorphic sorts. To address this, I had previously introduced a hack that limited the 'depth' of polymorphic tags to 1 (so we could have a serialized tag corresponding to the function type `List Int -> String`, but not `List (List Int) -> String`). 

This PR solves both issues by expanding the tag table with new function sorts only when they are encountered by the SMT serialisation mechanisms, and not trying to cover all their possible combinations.

> * How to register the need to generate an apply?
> * How to generate an apply that is necessary?

Technically, new tags are registered in the context by the SMT serialisation functions `smt2App`, `smt2Coerc`, `smt2Lam` and `smtLamArg` in `L.F.Smt.Serialize`. When `L.F.Smt.Interface.interactDecl'` is invoked, all new tags are written to the SMT script and moved to the corresponding level of the tag table stack, ensuring there are no duplicate definitions (which cause the SMT solver to crash) within the scope of a single `push`/`pop` bracket. `interactDecl'` itself is wrapped in 3 user-facing functions `smtAssertDecl`, `smtDefineEqn` and `smtDefineFunc`.

> * When to call `clearApplies`?

For some reason, `L.F.Solver.Monad.sendConcreteBindingsToSMT` seems to interfere with the process of removing the definitions from the tag table stack so they can be regenerated after the next addition, which causes the definitions in the SMT script and LF's knowledge of them to become desynchronised, resulting in 'unknown function/constant' errors. As a workaround, I've added the 'clearApplys'. This should be investigated further.

> * What is the relation between lam_arg and applies that causes lam_arg to be disabled without --higherorder?

The problem with `lam_arg` is that, unlike the other three symbols, which are generated once per tag, `lam_arg` is generated in batches. Currently, the size of these batches is set to 20 (see `L.F.Smt.Theories.maxLamArg`), which increases the size of the script somewhat. Ideally, we should only generate the `lam_arg` symbols where they are needed. As a heuristic for this, I've added the `ctxLams` flag, which is only enabled when `--higherorder` is. This seems to work on all available tests.

> Also, I'm guessing one change in this PR must be that applies were pregenerated once in a preamble, whereas now they might be produced more than once at the places where they are needed.

Yes, this is a consequence of the "lazy" approach: we essentially discard all newly generated symbols after a corresponding `pop` and reset the fresh tag number to that of the previous level (which is stored in the `ctxIxs` stack). This behaviour doesn't seem to cause any performance issues.

facundominguez:
> The concepts of 'apply', 'lam_arg', 'coerce' and 'lambda' are not new to the PR; 

Thanks Alex, I feel that is a good summary to approach the changes. Though I admit it will take me some time to study the code to recover how these concepts work. If this is not documented, and you have built an understanding of how they relate to each other, it could be helpful to explain them briefly next to the functions that introduce them or manipulate them. Otherwise, someone or myself will have to do that eventually.

The comments on the table of applies and how it is updated and cleaned could be next to environment or the context where it is kept.

> which is only enabled when --higherorder is. 

If `--higherorder` is disabled and the lam_arg symbols are missing, will we get an error message hinting that we should try enabling it?

Given my scarce availability to review at the moment, I don't object to merging first and reviewing later :shrug:

Let me know when you think it is good to merge, or perhaps @nikivazou should merge it.

nikivazou:
It looks good to me! 
Before merging, can you update the relevant LH pull request to make sure that nothing breaks there either?

clayrat:
> If `--higherorder` is disabled and the lam_arg symbols are missing, will we get an error message hinting that we should try enabling it?

Unfortunately not. An actual error message would come from the SMT solver complaining about the missing `lam_arg$XXX` definition, and I'm not sure how to instrument that. The only way I can think of to implement such a warning on the LF side is to add a flag check to the call to `smtLamArg`, but this would most likely impact performance in cases where HO reasoning is happening.

PR #753, opened 2025-05-21
elaboration

nikivazou:
Refine elaboration as required for https://github.com/ucsd-progsys/liquidhaskell/issues/2535

This edit required for the test https://github.com/ucsd-progsys/liquid-fixpoint/blob/2f3c896c8790738d2701ac109e14ee6525d7efc1/tests/pos/T753.fq 

because it needs the sort of the function in a function call.

### DIFF
--- a/src/Language/Fixpoint/SortCheck.hs
+++ b/src/Language/Fixpoint/SortCheck.hs
@@ -1359,13 +1359,13 @@ unify1 f e !θ !t (FVar !i)
   = unifyVar f e θ i t
 unify1 f e !θ t1 t2 
   | [FTC bg, tt1] <- unFApp t1
-  , bagConName == symbol bg 
+  , bagConName == symbol bg || setConName == symbol bg
   , [FTC arr, tt2, _] <- unFApp t2
   , arrayConName == symbol arr 
   = unify1 f e θ tt1 tt2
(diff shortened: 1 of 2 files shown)


nikivazou:
Can we merge?

facundominguez:
Thanks Niki! I don't oppose merging, but I think that without additional action it will be difficult for others to maintain the code.

It looks to me like the test `753.fq` is too large. Imagine it starts failing in the future, how would you make sense of all the constraints in there? If it comes from Liquid Haskell, I'd let it stay in Liquid Haskell.

Also, two artifacts would make maintaining this code easier:
1. A small test that rehearses the changes 
2. A comment explaining what the problem was with the old implementation and how the changes addresses it. The code could have it, or it could have a reference to the PR or issue where it is explained. Otherwise, I'd expect the insights to get lost in the flux of time.

nikivazou:
ok! I will try to minimize the test!

nikivazou:
Shall we merge?

facundominguez:
Thanks Niki! I removed the large `tests/pos/T753.fq` now that there is the smaller `tests/pos/T753a.fq`. But please, let me know if this is not what you want. I think we could merge when CI passes.

PR #754, opened 2025-06-27
add a pass for sort coercion in elaboration

clayrat:
This is a follow-up PR for https://github.com/ucsd-progsys/liquid-fixpoint/pull/753/. It adds a separate pass for sort coercion during elaboration and fixes the handling of casted atoms. This allows us to eliminate the ad hoc clauses in the unifier.


nikivazou:
Thanks!!! Looks good to me!

nikivazou (approved):

PR #756, opened 2025-07-23
Create nightly release workflow

nilehmann:
Add a workflow to publish "nightly releases" with binaries.

* A release is published every time a commit is pushed to `develop`.
* The release always points to the same tag, i.e., there's always a single nightly release.
* This is what the release looks like https://github.com/nilehmann/liquid-fixpoint/releases/tag/nightly.
* The workflow builds binaries for linux x86_64, macos x86_64, and macos aarch64.
* The workflow runs on PRs to build the binaries as a sanity check, but no release is published until merging. Building for macOS x86_64 is kind of slow (~25min), so I can modify the workflow to only run after merging if you want. Although there's some caching enabled for PRs, so this may be ok.

### DIFF
--- a/.github/actions/github-release/main.js
+++ b/.github/actions/github-release/main.js
@@ -1,0 +1,149 @@
+const core = require('@actions/core');
+const path = require("path");
+const fs = require("fs");
+const github = require('@actions/github');
+const glob = require('glob');
+
+function sleep(milliseconds) {
+  return new Promise(resolve => setTimeout(resolve, milliseconds));
+}
+
+async function runOnce() {
+  // Load all our inputs and env vars. Note that `getInput` reads from `INPUT_*`
+  const files = core.getInput('files');
+  const releaseName = core.getInput('name');
+  const tagName = core.getInput('tag_name');
+  const tagMessage = core.getInput('tag_message');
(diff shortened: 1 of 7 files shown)


.github/actions/github-release/main.js
```diff
@@ -0,0 +1,149 @@
...
+  try {
+    core.info(`> updating tag`);
+    await octokit.rest.git.updateRef({
+      owner,
+      repo,
+      ref: `tags/${tagName}`,
+      sha,
+      force: true,
+    });
+  } catch (e) {
```
facundominguez, line 57: Can it be tested here if the failure is due to the tag not being present?

facundominguez (approved):
Looks good to me in general. Though I'm not familiar with the API for actions.

> The workflow runs on PRs to build the binaries as a sanity check

Perhaps the linux job is not necessary to run on PRs as the build is done in another job already.

.github/actions/github-release/main.js
nilehmann: Good question. I copied this from rust-analyzer as is. Presumably yes, but this is hard to test outside of github actions and the documentation of octokit doesn't have much information regarding exceptions.

PR #758, opened 2025-08-08
Build windows binaries in nightly releases

nilehmann:
The main change in this PR is to add a windows binary to nightly releases, but there are a couple of extra general improvements

* ~Cache build artifacts based on `cabal freeze`~
* Compress assets before uploading them
* Cache dependencies based on plan.json

I tested the binary in an old laptop and it passes all Flux's tests

Example of how the release looks like https://github.com/nilehmann/liquid-fixpoint/releases/tag/nightly

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,39 +28,57 @@ jobs:
             artifact: fixpoint-x86_64-apple-darwin
           - os: macos-14
             artifact: fixpoint-aarch64-apple-darwin
+          - os: windows-2025
+            artifact: fixpoint-x86_64-pc-windows
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
 
       - name: Set up Haskell
+        id: setup-haskell
         uses: haskell-actions/setup@v2
         with:
           ghc-version: ${{ env.GHC_VERSION }}
           cabal-version: ${{ env.CABAL_VERSION }}
 
-      # Use cache in PRs, but do a clean build before releases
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -28,39 +28,57 @@ jobs:
...
         with:
           ghc-version: ${{ env.GHC_VERSION }}
           cabal-version: ${{ env.CABAL_VERSION }}
 
-      # Use cache in PRs, but do a clean build before releases
-      - name: Cache ~/.cabal/packages, ~/.cabal/store and dist-newstyle
+      - name: Freeze Cabal plan
+        run: |
+          cabal update
+          cabal freeze
```
facundominguez: Note that the documentation of haskell-actions/setup provides [examples](https://github.com/haskell-actions/setup/blob/main/docs/examples.md#model-cabal-workflow-with-caching) with `plan.json` instead of `cabal freeze`. `plan.json` does seem to be more precise.

nilehmann: That makes sense, thanks for the suggestion. I also noticed they only cache dependencies. Do you know any pros/cons of doing that?

facundominguez: I don't have strong opinions there. Caching dependencies allows to save the cache if the build of the local packages fails.

Caching the full build might save some extra work as long as incremental builds are reliable. I haven't needed to clean caches more than once or twice in many years of using Haskell.

nilehmann: I went with just caching dependencies.

PR #762, opened 2025-09-08
Shorten flags and add git-version

ranjitjhala:
I shortened the flag names so the `--help` renders more nicely (otherwise the text is weirdly squished into too few columns making it hard to read...) This may require matching updates to LH.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ Liquid Fixpoint
 ===============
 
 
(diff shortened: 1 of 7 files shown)


ranjitjhala:
@facundominguez you might want to take a look?

facundominguez:
I'm fine shortening the names.

The best solution would be to implement a different help formatting in `cmdargs` [[link]](https://hackage.haskell.org/package/cmdargs). So the documentation is not constrained by the length of flag names.

Another option would be using explicit names when configuring `cmdargs`, instead of renaming record fields (from the documentation: `{hello = def &= explicit &= name "foo"}`).

Other than that, I think the changes in this PR are tolerable.

ranjitjhala:
Thanks @facundominguez  -- I tried using the `explicit` but it seems to keep the old name still, oddly? Am merging for now, we can revert/fix if needed!

facundominguez:
I asked about changing the help format in cmdargs (https://github.com/ndmitchell/cmdargs/issues/72).

PR #763, opened 2025-10-14
Generalize bindings

pieter-bos:
Generalize bindings (e.g. `Symbol` in `ELam`, `ELet`) to a type variable `b`. Similarly to when `v` was added, rename `ExprV` to `ExprBV` with a new type parameter, and add a type alias `ExprV v = ExprBV Symbol v`. Also generalize `Subable` over variables.

This is to support [#2572](https://github.com/ucsd-progsys/liquidhaskell/pull/2572), where the refinements of refinement types are usually stored as a fixpoint `Expr`. Since `ExprBV` requires that `SubstV` is also generalized, I thought it best to consider it jointly with `Subable`.

Some wrinkles to consider:

* For `Subable` I looked at various options, but there's not a clear winner for me. This try is to add a `Variable` type family to `Subable`, which makes sense to me: a type should know the variables it stores, or it is probably a convenience instance that does nothing (e.g. `Subable ()`). For existing instances without requirements it can safely remain a `Symbol` by default, but I generalize the instances in most places here.

  For sites that require `Subable x` for some `x` it was previously guaranteed that the associated functions like `syms` accept / return Symbols, so it is a breaking change for such sites. 
  
  Another option is to add a new class that is generic over variable types (e.g. `SubableV v a`). We could leave it as a separate class, or connect it with `Subable` in various ways. (e.g. make `Subable` a requirement of `SubableV`, or change `Subable` to a constraint alias `Subable = SubableV Symbol`, or add default signatures to `Subable` that require `SubableV Symbol`).
* I don't think substitutions make sense unless `b ~ v`. Consequently I'm not sure whether `SubstBV` deserves to have a separate `b` and `v`. This is currently required because it is also stored as the substitution next to k-vars. 
  
  It would at least make more sense if `SubstBV` would key over `v` instead of `b`, as that's what we're substituting, but this makes it impossible to implement `Functor (SubstBV b)`, as `HashMap` is not a functor in its key. Given that we only apply substitutions when `b ~ v` we can also just write it off as an abuse of notation, but I thought it deserves some consideration.

---

This work has been supported by funding from the _Agentur für Innovation in der Cybersicherheit GmbH ([Cyberagentur](https://www.cyberagentur.de))_ as part of the [Clash Formal](https://clash-formal.org) project.

### DIFF
--- a/src/Language/Fixpoint/SortCheck.hs
+++ b/src/Language/Fixpoint/SortCheck.hs
@@ -286,7 +286,7 @@ elabFMap (PAtom r e1 e2)   = PAtom r (elabFMap e1) (elabFMap e2)
 elabFMap (PAll   bs e)     = PAll bs (elabFMap e)
 elabFMap (PExist bs e)     = PExist bs (elabFMap e)
 elabFMap (ECoerc a t e)    = ECoerc a t (elabFMap e)
-elabFMap (PKVar k (Su m))  = PKVar k (Su (elabFMap <$> m))
+elabFMap (PKVar k su)      = PKVar k (toKVarSubst (elabFMap <$> fromKVarSubst su))
 elabFMap e                 = e
 
 
@@ -333,7 +333,7 @@ elabFSetBagZ3 = go
     go (PAll   bs e)      = PAll bs (go e)
     go (PExist bs e)      = PExist bs (go e)
     go (ECoerc a t e)     = ECoerc a t (go e)
-    go (PKVar k (Su m))   = PKVar k (Su (go <$> m))
+    go (PKVar k su)       = PKVar k (toKVarSubst (go <$> fromKVarSubst su))
     go e                  = e
 
 -- | Reverse transformation of elabFSetBagZ3: converts array representations back to set/bag operations
@@ -412,7 +412,7 @@ unElabFSetBagZ3 = go
     go (PAll   bs e)      = PAll bs (go e)
     go (PExist bs e)      = PExist bs (go e)
     go (ECoerc a t e)     = ECoerc a t (go e)
-    go (PKVar k (Su m))   = PKVar k (Su (go <$> m))
+    go (PKVar k su)       = PKVar k (toKVarSubst (go <$> fromKVarSubst su))
     go e                  = e
 
 
@@ -435,7 +435,7 @@ elabSorts ef (PAtom r e1 e2)   = PAtom r (elabSorts ef e1) (elabSorts ef e2)
 elabSorts ef (PAll   bs e)     = PAll bs (elabSorts ef e)
 elabSorts ef (PExist bs e)     = PExist bs (elabSorts ef e)
 elabSorts ef (ECoerc s1 s2 e)  = ECoerc (coerceSort ef s1) (coerceSort ef s2) (elabSorts ef e)
-elabSorts ef (PKVar k (Su m))  = PKVar k (Su (elabSorts ef <$> m))
+elabSorts ef (PKVar k su)      = PKVar k (toKVarSubst (elabSorts ef <$> fromKVarSubst su))
 elabSorts _ e                 = e
 
 --------------------------------------------------------------------------------
@@ -502,7 +502,7 @@ elabApply env = go
     step e@EApp {}        = go e
     step (ELam b e)       = ELam b       (go e)
     step (ECoerc a t e)   = ECoerc a t   (go e)
-    step (PKVar k (Su m)) = PKVar k (Su (go <$> m))
+    step (PKVar k su)     = PKVar k (toKVarSubst (go <$> fromKVarSubst su))
     step e@ESym{}         = e
     step e@ECon{}         = e
     step e@EVar{}         = e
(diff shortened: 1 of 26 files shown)


facundominguez:
Thanks @pieter-bos!

Maybe it could help to not use `SubstBV` for kvars, so the functor requirement does not become an impediment. Alternatively, `SubstBV` could be changed to not use a `HashMap`. Then we would have to measure if the `HashMap` makes much of an impact in any of the tests.

I don't have much opinion on which way to go with `Subable`. If the current approach works, I think we could use it.

pieter-bos:
Yes, good plan :). I've refactored `ExprBV` to contain a `KVarSubst` which stores the mapping as a list instead, with the standard way to access it [here](https://github.com/ucsd-progsys/liquid-fixpoint/pull/763/files#diff-3ba29331d2bb3ab24ac015fd6b71ceb741c0af4886bd2c0d175c6446629adc1eR303-R307) via a map so there is no confusion about maps (lists) with duplicate keys.

src/Language/Fixpoint/Types/Refinements.hs
```diff
@@ -417,7 +453,7 @@ exprSymbolsSet = go
     go (PIff p1 p2)       = gos [p1, p2]
     go (PImp p1 p2)       = gos [p1, p2]
     go (PAtom _ e1 e2)    = gos [e1, e2]
-    go (PKVar _ (Su su))  = HashSet.unions $ map exprSymbolsSet (M.elems su)
+    go (PKVar _ su)       = HashSet.unions $ map exprSymbolsSet (M.elems $ fromKVarSubst su)
```
facundominguez:
```suggestion
    -- Convert the substitution to a hash map to eliminate duplicate
    -- variable entries.
    go (PKVar _ su)       = HashSet.unions $ map exprSymbolsSet (M.elems $ fromKVarSubst su)
```
Is this why the substitution is converted to a HashMap?

facundominguez:
This is looking good to me. Could you compare CPU time in Liquid Haskell for these changes using these steps [[link]](https://github.com/ucsd-progsys/liquidhaskell?tab=readme-ov-file#how-to-create-performance-comparison-charts)?

facundominguez:
I just updated the instructions in the readme to include removing all artifacts before measuring time.

ranjitjhala:
hi @facundominguez sorry I seem to have missed this, do you know the status here? Should we try to resolve conflicts and merge?

facundominguez:
IIUC, Liquid Haskell still needs to be updated to cope with these changes. The author is looking into that before this PR is finished.

ranjitjhala:
Thanks @pieter-bos and @facundominguez !

pieter-bos:
Hi again! I've made good progress on the changes required in `liquidhaskell`, so I was able to do a profiling run. Note that this thus includes the changes [here](https://github.com/ucsd-progsys/liquidhaskell/compare/ucsd-progsys:ca4e61f...QBayLogic:6974759), which I still plan to clean up a bit and submit. 

bot
top

The diff is a bit of a mixed bag, I don't really have a hypothesis to explain it well. I did run the [before] situation once more, there was no difference between the [before] runs. The runs are for these commits:

 - before [[liquidhaskell]](https://github.com/ucsd-progsys/liquidhaskell/commit/ca4e61f517360674c99eb580330598b48abcc69d) [[liquid-fixpoint]](https://github.com/ucsd-progsys/liquid-fixpoint/commit/2b74a261d1905cf2f4e89969120815755ea613ba)
 - after [[liquidhaskell]](https://github.com/QBayLogic/liquidhaskell/commit/69747593ac25a5102026f91fd0af1516328790d9) [[liquid-fixpoint]](https://github.com/QBayLogic/liquid-fixpoint/commit/f74f0d75c3dd8080fa69a8275c817d20a8d6b282)

src/Language/Fixpoint/Types/Environments.hs
```diff
@@ -111,74 +116,74 @@ splitByQuantifiers (BE i bs) ebs = ( BE i $ M.filterWithKey (\k _ -> notElem k e
...
 unionSEnv' (SE m1) (SE m2)    = SE (M.union m1 m2)
 
 {-# SCC lookupSEnvWithDistance #-}
-lookupSEnvWithDistance :: Symbol -> SEnv a -> SESearch a
+lookupSEnvWithDistance :: Binder b => b -> SEnvB b a -> SESearchB b a
 lookupSEnvWithDistance x (SE env)
   = case M.lookup x env of
      Just z  -> Found z
-     Nothing -> Alts $ symbol <$> alts
+     Nothing -> Alts $ alts
```
facundominguez:
```suggestion
     Nothing -> Alts alts
```

facundominguez (commented):
LGTM!

Maybe we can merge despite of the mixed performance results. I'll double check the comparisons a bit later.

facundominguez:
Measuring with `--measure-timings-j1` shows that the performance differences are small enough. :tada:

bot-wc
top-wc

In order to merge this, we would need also a PR in Liquid Haskell with the update for these changes.

I made a mistake that could distort the results of `--measure-timings`, which I'm fixing in https://github.com/ucsd-progsys/liquidhaskell/pull/2624.

src/Language/Fixpoint/SortCheck.hs
```diff
@@ -412,7 +412,7 @@ unElabFSetBagZ3 = go
     go (PAll   bs e)      = PAll bs (go e)
     go (PExist bs e)      = PExist bs (go e)
     go (ECoerc a t e)     = ECoerc a t (go e)
-    go (PKVar k (Su m))   = PKVar k (Su (go <$> m))
+    go (PKVar k su)       = PKVar k (toKVarSubst (go <$> fromKVarSubst su))
```
facundominguez: Maybe this pattern could be replaced with `mapKVarSubst go`. It would not delete duplicates, but perhaps that is not necessary.

src/Language/Fixpoint/Types/Substitutions.hs
```diff
@@ -193,42 +226,41 @@ instance Subable Expr where
...
             su' = substExcept su (map fst bs)
         p ->
           p
 
-removeSubst :: Subst -> Symbol -> Subst
+removeSubst :: (Eq v, Hashable v) => SubstV v -> v -> SubstV v
 removeSubst (Su su) x = Su $ M.delete x su
 
+class Refreshable v where
+  candidates :: v -> [v]
```
facundominguez, line 255:
```suggestion
-- | Variable names for which we can propose variations to avoid name captures
class Refreshable v where
  -- | Variations of a variable name. They must contain at least a fresh
  -- name in the contexts where @candidates@ is used. 
  candidates :: v -> [v]
```

src/Language/Fixpoint/SortCheck.hs
pieter-bos: Oh, good point. My idea was to have the invariant that `KVarSubst` does not contain duplicate keys, by not exporting its constructor and making `KVarSubst`s only via `toKVarSubst`. If that's true it should be fine to use `mapKVarSubst` in this kind of case.

PR #764, opened 2025-10-14
Extend/Fix the Solution Printer

ranjitjhala:
This PR makes a few related changes

- Make the `--json` also dump out the `Solution` as part of the result,
- Add the "params" of each `KVar` as a part of the `Solution`,

Note that this requires 

1. *not* quantifying out the `xts` that come from the params of the `KVar` in `cubePred`
2. *not* removing the equalities due to the above (which are now "free" in the solution).

@facundominguez would be most grateful for any comments if you have time!

So now, we get

```
$ cabal run fixpoint -- tests/horn/pos/sum-rec.smt2 --save
$ more tests/horn/pos/.liquid/sum-rec.smt2.fqout

Solution:
$k1[$k1##0 : int, $k1##1 : int] := $k1##0 >= 0
                                   && $k1##0 >= $k1##1


Non-cut kvars:

```

and 

```
$ cabal run fixpoint -- tests/horn/pos/mod00.smt2 --save
...
$ more tests/horn/pos/.liquid/mod00.smt2.fqout

Solution:



Non-cut kvars:

$k0[$k0##0 : int] := exists [a1##2 : int]
                       . $k0##0 == a1##2
                         && a1##2 == 10
                     || exists [a0##1 : int]
                          . $k0##0 == a0##1
                            && a0##1 == 4
```

### DIFF
--- a/src/Language/Fixpoint/Solver.hs
+++ b/src/Language/Fixpoint/Solver.hs
@@ -33,7 +33,6 @@ import qualified Data.Text.Lazy.IO                as LT
 import qualified Data.Text.Lazy.Encoding          as LT
 import           System.Exit                        (ExitCode (..))
 import           System.Console.CmdArgs.Verbosity   (whenNormal, whenLoud)
-import           Text.PrettyPrint.HughesPJ          (render)
 import           Control.Monad                      (mplus, when)
 import           Control.Exception                  (catch)
 import           Language.Fixpoint.Solver.EnvironmentReduction
@@ -60,6 +59,8 @@ import           Language.Fixpoint.Solver.Instantiate (instantiate)
 import           Control.DeepSeq
 import qualified Data.ByteString as B
 import Data.Maybe (catMaybes, mapMaybe)
+import Data.Ord (comparing)
+import qualified Text.PrettyPrint.HughesPJ as PJ
 
 ---------------------------------------------------------------------------
 -- | Solve an .fq file ----------------------------------------------------
@@ -86,7 +87,7 @@ resultExitCode cfg r = do
     jStr    = LT.decodeUtf8 . encode $ r
     stat    = resStatus $!! r
     eCode   = resultExit . resStatus
-    statStr = render . resultDoc
+    statStr = PJ.render . resultDoc
 
 ignoreQualifiers :: Config -> FInfo a -> FInfo a
 ignoreQualifiers cfg fi
@@ -207,7 +208,7 @@ type ErrorMap a = HashMap.HashMap SrcSpan a
 findError :: ErrorMap a -> Error1 -> Maybe ((Integer, a), Maybe String)
 findError m e = do
   ann <- HashMap.lookup (errLoc e) m
-  let str = render (errMsg e)
+  let str = PJ.render (errMsg e)
   return ((-1, ann), Just str)
 
 -- The order is important here: we want the "binders" to get the "precedence"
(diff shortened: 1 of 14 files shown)


src/Language/Fixpoint/Solver.hs
```diff
@@ -340,52 +353,55 @@ simplifyResult res =
...
-    --
-    isVarToDrop fvs (EApp (EVar "cast_as_int") ei) = isVarToDrop fvs ei
-    isVarToDrop fvs (EVar s)
-      | elem [s] fvs = Just s
-    isVarToDrop _fvs _ = Nothing
-
-simplifyKVar e = e
+    go (POr es) = POr $ map go es
+    go (PExist bs e@(PAnd es)) =
+      let fvs = L.group $ L.sort $ freeVars ++ collectFreeVarOccurrences e
```
facundominguez: :wave: Maybe we should only eliminate bindings of `bs`, that is existentially quantified. That way, there would not be risk of removing other variables.

How about we define
```Haskell
    fvs =
      [ g
      | g <- L.group $ L.sort $ collectFreeVarOccurrences e
      , v:_ <- g
      , elem v (map fst bs)
      ]
```
?

If I'm missing the point of the change, perhaps I need an example of a formula that the old `simplifyKVar` is "simplifying" wrong.

ranjitjhala:
Oops silly me - I was going to include the "old" versions of the `fqout` in the PR message but forgot, sorry!

Here's what you get in `develop` (old) for those two examples

```
$ more tests/horn/pos/.liquid/mod00.smt2.fqout

Solution:



Non-cut kvars:

$k0 := exists [a1##2 : int]
         . a1##2 == 10
       || exists [a0##1 : int]
            . a0##1 == 4
```

The above solution for `$k0` has "lost" the key information -- the predicate as printed above is just a tautology.

What we want instead is the solution printed above, namely

```
$k0[$k0##0 : int] := exists [a1##2 : int]
                       . $k0##0 == a1##2
                         && a1##2 == 10
                     || exists [a0##1 : int]
                          . $k0##0 == a0##1
                            && a0##1 == 4
```

which says (morally) that `$k(v) := v == 10 || v ==4`.

Earlier (that is in `develop`, the `cubePred` function was returning something like

```
                  exists [$k0##0 : int, a1##2 : int]
                       . $k0##0 == a1##2
                         && a1##2 == 10
                     || exists [a0##1 : int]
                          . $k0##0 == a0##1
                            && a0##1 == 4
```

where `$k0##0` the name given to the "parameter" of the horn variable `$k0`, was *existentially quantified out*,
as it is in `xts`, and so the subsequent `simplifyKVar`'s heuristics went ahead and eliminated the equality.

The main changes now are to:

1. Not existentially quantify out,
2. Pass in those variables as extra "free variables" that should not be eliminated by `simplifyKVar`. Does that help?

src/Language/Fixpoint/Solver.hs
ranjitjhala: Ah I see your approach is very clean; I'll remove the "freeVars" thing that I did!

facundominguez:
Thanks, that helps me check my understanding. I'm proposing that instead of preserving equalities about the parameters, we preserve equalities about any variable that is not an existentially quantified variable, if that makes sense to you.

In this example,
```
$k0[$k0##0 : int] := exists [a1##2 : int]
                       . $k0##0 == a1##2
                         && a1##2 == 10
                     || exists [a0##1 : int]
                          . $k0##0 == a0##1
                            && a0##1 == 4
```
we still won't eliminate `$k0##0 == a1##2` or `$k0##0 == a0##1`, but instead of preserving them because `$k0##0` is a parameter, we preserve them because `$k0##0` is neither bound by `exists [a1##2 : int]` nor by `exists [a0##1 : int]`.

What do you think?

ranjitjhala:
Yes, I like your approach -- have made that change locally and will push shortly.

PR #766, opened 2025-10-15
Revise documentation of simplifyKVar

facundominguez:
Follow up to #764.

This uses a more direct representation of free variable occurrences, and makes the surrounding comments a bit more precise.


ranjitjhala (approved):
This looks great, thanks @facundominguez !

PR #770, opened 2025-10-19
Various changes to support extracting solutions

ranjitjhala:
1. Add a `--fullsolution` flag that disables simplifications, name tidying
2. Add `unElab` passes to revert the lowering of names into Z3 (e.g. for sets/maps)

The current `simplifyKVar` produces incorrect solutions, where there are odd errors with 
- name mangling when we use `tidySymbol` (only applies to expr-body, incorrectly to binders)
- non-equality predicates (e.g. bool) which don't quite fit the equality-occurrence based simplification heuristics.

So I added a `--fullsolution` that only GC's totally dead exists-binds and does nothing else, 
except that we "reverse" the elaboration done to lower to Z3's `map` based operators (as fixpoint 
clients are unaware of those, and only understand the public theory-symbols.)

### DIFF
--- a/src/Language/Fixpoint/Solver.hs
+++ b/src/Language/Fixpoint/Solver.hs
@@ -328,7 +328,7 @@ simplifyResult :: Result a -> Result a
 simplifyResult res =
     res
       { resSolution = HashMap.map simplifyKVar' (resSolution res)
-      , resNonCutsSolution = HashMap.map simplifyKVar' (resNonCutsSolution res)
+      , resNonCutsSolution = HashMap.map simplifyKVar' (tracepp "pre-simplify" $ resNonCutsSolution res)
       }
   where
     simplifyKVar' = unElab . simplifyKVar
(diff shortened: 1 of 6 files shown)


src/Language/Fixpoint/Solver.hs
```diff
@@ -324,14 +324,36 @@ saveSolution cfg res = when (save cfg) $ do
...
+
+simplifyKVarTrivial :: Expr -> Expr
+simplifyKVarTrivial = go
+  where
+    go (POr es)      = POr (go <$> es)
+    go (PAnd es)     = PAnd (go <$> es)
+    go (PExist bs e) = pExist bs' e
+      where
+        fvs = collectFreeVarOccurrences e
+        bs' = filter (\(b, _) -> b `elem` fvs) bs
```
facundominguez: Another alternative
```suggestion
    go (PExist bs e) = pExist bs' e
      where
        fvs = exprSymbolsSet e
        bs' = filter (\(b, _) -> b `HS.member` fvs) bs
```

src/Language/Fixpoint/Solver.hs
```diff
@@ -348,27 +370,37 @@ simplifyResult res =
...
           occursExactlyOnce [_] = True
           occursExactlyOnce _   = False
+          occursMoreThanOnce (_:_:_) = True
+          occursMoreThanOnce _ = False
 
           esv = map (isUniqueEq singleOccurrenceBindings) es
           removed = mapMaybe fst esv
-          needed = singleOccurrenceBindings L.\\ removed
+          needed = (singleOccurrenceBindings L.\\ removed) ++ multipleOccurrenceBindings
           bs' = filter ((`elem` needed) . fst) bs
```
facundominguez, line 382: 👍 Oh, I see how this was wrong. Would this simpler version work?
```suggestion
          bs' = filter ((`notElem` removed) . fst) bs
```

facundominguez:
> Add a --fullsolution flag that disables simplifications, name tidying

Maybe users could have an example in the documentation about when to enable `--fullsolution`. I myself, don't know yet when and how the simplification and tidying can go wrong.

That said, there is no place for documenting liquid-fixpoint yet. Perhaps the README would do for the moment?

src/Language/Fixpoint/Solver.hs
facundominguez: hm, that doesn't eliminate bindings that aren't used. Maybe your version is to be preferred.

ranjitjhala: Yes, that doesn't remove the unused bindings AFAICT.

ranjitjhala:
@facundominguez I think you're right! I _thought_ there was some issue with the tidying (and _should_ have kept the test around) but now I can't seem to replicate it :-) -- that is, the regular `tidySolution` seems to be doing the right thing, so I'll remove the `--fullsolution` flag. Thanks!

ranjitjhala:
Otherwise ok to merge?

facundominguez (approved):
LGTM!

PR #772, opened 2025-10-23
Fixes for the `--scrape` heuristic

ranjitjhala:
This PR fixes the scrape heuristic so that

1. It also considers qualifiers over global constants and ADT (and perhaps other theory) symbols,
2. Fixes the qualifier instantiation so that it allows for the above "global" symbols.

This could cause performance problems because now we're considering potentially _more_ qualifiers, 
so perhaps should be guarded by a flag, but I'd like to benchmark LH before and after this change 
to see if a flag is actually needed?


src/Language/Fixpoint/Solver/Sanitize.hs
```diff
@@ -160,11 +161,11 @@ eliminateEta cfg si
...
-theoryEnv :: Config -> F.GInfo c a -> F.SEnv F.TheorySymbol
-theoryEnv cfg si
-  =  Thy.theorySymbols (Cfg.solver cfg)
-  <> Thy.theorySymbols (F.defns si)
-  <> Thy.theorySymbols (F.ddecls si)
+-- theoryEnv :: Config -> F.GInfo c a -> F.SEnv F.TheorySymbol
+-- theoryEnv cfg si
+--   =  Thy.theorySymbols (Cfg.solver cfg)
+--   <> Thy.theorySymbols (F.defns si)
+--   <> Thy.theorySymbols (F.ddecls si)
```
facundominguez:
```suggestion
```

This comment can be removed, I guess. `theoryEnv` is moved to the SortCheck module.

ranjitjhala:
Thanks @facundominguez !

Unfortunately, looks like this "improvement" really hurts LH ... so I'll have to think a bit about how to gate it...

filtered

src/Language/Fixpoint/Solver/Solution.hs
```diff
@@ -56,7 +56,7 @@ init cfg si ks_ = Sol.fromList symEnv mempty keqs [] mempty ebs xEnv
     qs_        = F.quals si
     ws         = [ w | (k, w) <- M.toList (F.ws si), not (isGWfc w), k `S.member` ks ]
     ks         = {- trace ("init-ks-size" ++ show (S.size ks_)) $ -} ks_
-    genv       = instConstants si
+    genv       = So.globalEnv cfg si <> instConstants si
```
ranjitjhala: I suspect this is the change that allows MANY more qualifiers into the initial solution, and hence should be gated...

ranjitjhala:
@facundominguez after adding the flag, I get this graph in LH, do you think its acceptable? I'm guessing the red "slowdown" is mostly noise?

filtered

facundominguez:
> @facundominguez after adding the flag, I get this graph in LH, do you think its acceptable?

Yes, I think so. Some improve and some worsen, but nothing that I'd say is extreme. Here are the charts from my run. This is CPU time without specifying any flags.

top
bot

tests/horn/pos/scrape02.smt2
```diff
@@ -0,0 +1,109 @@
+(fixpoint "--scrape=both")
```
facundominguez, line 1: These tests strike me as too large for debugging if they ever stop passing.

Could we have smaller tests that still demonstrate that qualifiers are scraped from where they should?

ranjitjhala:
Yes very good suggestion - let me shrink those tests!

- Ranjit.

PR #773, opened 2025-10-25
Document some more functions and eliminate redundant cut kvar computation

facundominguez:
Besides the documentation additions, this PR removes the computation of cut kvars done immediately after PLE.

As PLE doesn't change the dependency graph of kvars, it is useless to recompute the cut set. Moreover, the result of the kvars computation is immediately discarded.

### DIFF
--- a/src/Language/Fixpoint/Solver/Solve.hs
+++ b/src/Language/Fixpoint/Solver/Solve.hs
@@ -99,8 +99,7 @@ doInterpret cfg fi0 subcIds = do
   where
     update' fi ss = ss{ssBinds = F.bs fi'}
       where
-        fi' = (siQuery sI) {F.hoInfo = F.HOI (C.allowHO cfg) (C.allowHOqs cfg)}
-        sI  = solverInfo cfg fi
+        fi' = fi {F.hoInfo = F.HOI (C.allowHO cfg) (C.allowHOqs cfg)}
 
 {-# SCC doPLE #-}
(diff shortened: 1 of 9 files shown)


facundominguez:
Tests continue to pass in Liquid Haskell.

src/Language/Fixpoint/Solver/Solve.hs
```diff
@@ -110,8 +109,7 @@ doPLE cfg fi0 subcIds = do
   where
     update' fi ss = ss{ssBinds = F.bs fi'}
       where
-        fi' = (siQuery sI) {F.hoInfo = F.HOI (C.allowHO cfg) (C.allowHOqs cfg)}
-        sI  = solverInfo cfg fi
+        fi' = fi {F.hoInfo = F.HOI (C.allowHO cfg) (C.allowHOqs cfg)}
```
facundominguez: Any hint about why `solverInfo` was needed here?

The only effect I could identify is dropping constraints from SInfo where they use kvars and there are no cut-kvars in the RHS, which has no obvious purpose.

ranjitjhala: Hmm TBH I'm not sure, it seems like it might be running the `eliminate` stuff (to compute `sHyp` etc AFTER the PLE instantiation is run?)

facundominguez: It seems the code was introduced in https://github.com/ucsd-progsys/liquid-fixpoint/pull/462, but unfortunately it doesn't provide motivation for this aspect.

PR #775, opened 2025-10-27
Fix order of kvar outputs for > 10 args

cole-k:
It's lexicographic ordering...

Not sure how tests work here, but if there are tests for the pretty printer I have a test file I can share with expected output.

# Before

```
$k0[$k0##0 : int,
    $k0##1 : int,
    $k0##10 : int,
    $k0##2 : int,
    $k0##3 : int,
    $k0##4 : int,
    $k0##5 : int,
    $k0##6 : int,
    $k0##7 : int,
    $k0##8 : int,
    $k0##9 : int] :=  ...
```

# After

```
$k0[$k0##0 : int,
    $k0##1 : int,
    $k0##2 : int,
    $k0##3 : int,
    $k0##4 : int,
    $k0##5 : int,
    $k0##6 : int,
    $k0##7 : int,
    $k0##8 : int,
    $k0##9 : int,
    $k0##10 : int] :=  ...
```

# N.B.

You could argue that downstream consumers of these solutions should parse the arguments and reorder them since we do number them, but that seems cruel. Better to keep the order as intended.

### DIFF
--- a/src/Language/Fixpoint/Solver.hs
+++ b/src/Language/Fixpoint/Solver.hs
@@ -321,7 +321,7 @@ saveSolution cfg res = when (save cfg) $ do
     where
       scopedRender = PJ.render . PJ.vcat . map ncDoc . scoped
       scoped sol = [ (k, scope k, e) | (k, e) <- HashMap.toList sol]
-      scope k = L.sortBy (comparing fst) $ HashMap.lookupDefault [] k $ resSorts res
+      scope k = map (\(_bid, name, sort) -> (name, sort)) $ L.sortBy (comparing (\(bid, _, _) -> bid)) $ HashMap.lookupDefault [] k $ resSorts res
       ncDoc (k, xts, e) = PJ.hsep [ pprint k PJ.<> pprint xts, ":=", pprint e ]
 
 simplifyResult :: Config -> Result a -> Result a
--- a/src/Language/Fixpoint/Solver/Solve.hs
+++ b/src/Language/Fixpoint/Solver/Solve.hs
@@ -177,14 +177,14 @@ tidyResult _ r = r
 tidySolution :: F.FixSolution -> F.FixSolution
 tidySolution = fmap tidyPred
 
-tidyBind :: (F.Symbol, F.Sort) -> (F.Symbol, F.Sort)
-tidyBind (x, t) = (F.tidySymbol x, t)
+tidyBind :: (F.BindId, F.Symbol, F.Sort) -> (F.BindId, F.Symbol, F.Sort)
+tidyBind (i, x, t) = (i, F.tidySymbol x, t)
 
 tidyPred :: F.Expr -> F.Expr
 tidyPred =  go
   where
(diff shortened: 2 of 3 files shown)


cole-k:
It seems like there might not be a 1-1 relation between the `BindId` and the actual order, since I get this as an output

```
kvar k4 := (lambda (($k4##0 Int)
         ($k4##5 Int)
         ($k4##6 Int)
         ($k4##7 Int)
         ($k4##8 Int)
         ($k4##9 Int)
         ($k4##10 Int)
         ($k4##11 Int)
         ($k4##12 Int)
         ($k4##13 Int)
         ($k4##14 Int)
         ($k4##1 Int)
         ($k4##2 Int)
         ($k4##3 Int)
         ($k4##4 Int)) (exists
```

I don't understand what's going on here, this isn't even a lexicographical ordering failure. And I'm going to trust that this is also not the right order, but I suppose I could check the constraint again to be sure.

cole-k:
The args are named over here https://github.com/cole-k/liquid-fixpoint/blob/ck-fix-kvar-solution-args-order/src/Language/Fixpoint/Horn/Info.hs#L144

```haskell
kvInfo :: (F.PPrint a) => F.BindEnv a -> H.Var a -> (F.BindEnv a, KVInfo a)
kvInfo be k       = (be', KVInfo k (Misc.fst3 <$> xts) wfc)
  where
    -- make the WfC
    wfc           = F.WfC wenv wrft  (H.hvMeta k)
    wenv          = F.fromListIBindEnv ids
    wrft          = (x, t, F.KV (H.hvName k))
    -- add the binders
    (be', ids)    = L.mapAccumL insertBE be xts'
    ((x,t,_), xts') = Misc.safeUncons "Horn var with no args" xts
    -- !!! ARGS NAMED HERE !!!
    -- make the parameters
    xts           = [ (hvarArg k i, t', a) | (t', i) <- zip (H.hvArgs k) [0..] ]
    a             = H.hvMeta k
```

and they are numbered in the order that they should be provided

```haskell
hvarArg :: H.Var a -> Int -> F.Symbol
hvarArg k i = F.hvarArgSymbol (H.hvName k) i
```

This seems to indicate that:
1. The **names** of the kvar args are numbered in the order they should be passed (e.g. `$k4##10` should be given before `$k4##11` and after `$k4##9`; it should be the 11th argument).
2. The `BindId`s of these args appear to **not necessarily** reflect this order.

Unless I'm missing some place where they are output and their order is rearranged.

ranjitjhala:
Yes that's definitely odd. Whats the .smt2 file that tickles this behavior?

cole-k:
[flux_demo.neural-{impl#0}-backward.smt2.txt](https://github.com/user-attachments/files/23173921/flux_demo.neural-.impl.0.-backward.smt2.txt)

Had to rename it so that GH would allow an upload.

cole-k:
Resolved by https://github.com/ucsd-progsys/liquid-fixpoint/pull/778

PR #779, opened 2025-10-29
Fix KVar args order in json output

cole-k:
#778 missed the spot where this is output when `fixpoint` is run with `--json`.


ranjitjhala:
Oops thanks for catching that! Will merge once CI finishes

ranjitjhala:
@facundominguez any idea what is going on with these somewhat random failures only with the CVC5 and only for some random tests (which seem different every time?)

facundominguez:
:wave: I haven't seen them before. Do they always end with the `<<loop>>` message?

PR #781, opened 2025-10-29
Explain a bit more refine and solverInfo

facundominguez:


src/Language/Fixpoint/Solver/Eliminate.hs
```diff
@@ -21,6 +21,17 @@ import           Language.Fixpoint.Solver.Sanitize
...
+-- Computes the set of cut and non-cut kvars, then initializes the solutions of
+-- the non-cut KVars (in the sHyp field)
+--
+-- The concept of cut KVars comes from the FUSION algorithm described in:
+--
+-- "Local Refinement Typing", ICFP 2017, https://ranjitjhala.github.io/static/local_refinement_typing.pdf
+--
+-- Note though, that the implementation here is not trying to profit from the
+-- structure of the lexical scope of the program to reduce the size of the kvar
+-- solutions.
```
facundominguez: I guess this is because fixpoint queries have constraints already flattened.

Is this implemented in the Horn solver though? I couldn't find the spot where shared hypotheses are kept out of constraints when computing the solutions (on a first glance).

ranjitjhala: Hi @facundominguez -- apologies for the spottiness of this documentation and thanks for trying to improve it!
So the place we compute the information is here

https://github.com/ucsd-progsys/liquid-fixpoint/blob/a0e46db38d1a8822e57ab9050df0bdf5b5f0b4fc/src/Language/Fixpoint/Types/Solutions.hs#L230-L231

this field, whose value is computed in this function `kvScopes` 
https://github.com/ucsd-progsys/liquid-fixpoint/blob/a0e46db38d1a8822e57ab9050df0bdf5b5f0b4fc/src/Language/Fixpoint/Solver/Eliminate.hs#L43-L44

tracks the set of binders that occurs in ALL the places that the kvar is used, we can leave these binders "free" in the solution of the kvar, i.e. no need to existentially quantify them away (as is done for the OTHER binders that might appear in the solution for a KVar). 

This business -- and why it is important -- is somewhat explained in Sec 2.4 of this paper 
https://ranjitjhala.github.io/static/local_refinement_typing.pdf

The "Horn" stuff is indeed "flattened" into the `FInfo` constraints, but we "recover" the lexical scope by computing the `kvScopes`. Does that help?

ranjitjhala: > I couldn't find the spot where shared hypotheses are kept out of constraints when computing the solutions (on a first glance).

This happens over here in `cubePred` https://github.com/ucsd-progsys/liquid-fixpoint/blob/a0e46db38d1a8822e57ab9050df0bdf5b5f0b4fc/src/Language/Fixpoint/Solver/Solution.hs#L443 

Specifically we "exclude" the bindings in `kvScopes k` from the predicate returned by `cubePred` for that particular occurrence of `k` as a "HEAD" constraint.

ranjitjhala:
Terrific, thanks!!!

src/Language/Fixpoint/Solver/Eliminate.hs
facundominguez: Ah, just the pointers I needed. Thanks!

I contributed some amendments. There is a precondition on the input of the .fq queries though: The subtyping constraints need to come from flattening some NNF query or lexical structure. Otherwise, LF would not find solutions to the following constraints:
```
c1: k1 => k2
c2: p => k1
c3: q => k2
```
This is because `k1` will only be allowed to use `p` and not `q`, since `q` doesn't occur in the constraints `c1` and `c2` where `k1` occurs.

ranjitjhala: There is a precondition -- its written down in the comments somewhere --   let me see if I can find it, that says the "common" environments of any two flattened constraints is a prefix of their entire environment. So given two constraints `cA` and `cB`, 
- IF `b1` < `b2` are BOTH in the environment of `cA` and `cB` 
- THEN for all `b1 < b < b2`, b in env of `cA` iff b in env `cB`.                                            

The case you outline is actually ok, because in that case, the disallowed bit (that is not shared) will simply be existentially quantified out -- i.e. you would "solve" `k1` to `exists x,y,z. p`.

facundominguez: :+1: This precondition came up before [[link]](https://github.com/ucsd-progsys/liquid-fixpoint/issues/496#issuecomment-937395718).

I don't think it is written in the source code. It would make a fine lint check to insert at a few key places.

facundominguez: Oh, here it is https://github.com/ucsd-progsys/liquid-fixpoint/blob/a6907c0d938afcb583e8f2994b5cb0111cd96be9/src/Language/Fixpoint/Solver/Instantiate.hs#L71-L77

PR #784, opened 2025-10-31
Remove the call to refine after PLE

facundominguez:
PLE cannot influence purposefully the solutions to kvars, as it ignores constraints with kvars.

It could accidentally affect the constraints that have kvars when it modifies bindings that also happen to be used in these constraints. But that is a behavior that we should stop as it makes constraint verification more unpredictable. #496 discusses a solution that does not need to modify bindings.


facundominguez:
All tests pass in Liquid Haskell

PR #786, opened 2025-11-03
Only recheck previously failed constraints after PLE

facundominguez:
Before this PR, all constraints would be checked, even those that were already known to pass before PLE.

This PR also removes the `ssBinds` field of `SolverState`. Passing the value explicitly makes clearer what functions produce it and consume it.


facundominguez:
All tests pass in Liquid Haskell

PR #788, opened 2025-11-03
Remove old ples

facundominguez:
After this PR only two PLE implementations remain: the interpreter and the regular SMT based.


ranjitjhala:
Godspeed @facundominguez !!!

PR #789, opened 2025-11-05
Remove PGrad and supporting code

### LINKED ISSUE #787: Remove PGrad from Expr
I checked today the uses of PGrad and I couldn't find any tests of it in either Liquid Haskell or Liquid Fixpoint. Liquid Haskell, in particular, never seems to insert this data constructor. Is it ok to remove it?

facundominguez:
Closes #787. The main motivation to remove this code is that it has no tests to prevent it from bit-rotting.


facundominguez:

PR #790, opened 2025-11-05
Second wave of deadcode removal after PGrad removal

facundominguez:
Follow up to #789


facundominguez:
All tests still pass in Liquid Haskell.

Issue #780, opened 2025-10-29
Substitute KVars in constraints for PLE


facundominguez:
If the LHS of a constraint has KVars, PLE ignores it. This is unfortunate because unresolved constraints can have KVars in this position.

It should be feasible to change liquid-fixpoint so it replaces these KVars with their best known solutions before feeding them to PLE. In this way, PLE has a chance to discover unfoldings that are needed by these constraints.

There is a fine example of this bug at work in https://github.com/ucsd-progsys/liquidhaskell/issues/2582

--------
### Implementation notes

There is a function `lhsPred` that converts the LHS of a constrain into a predicate with no KVars. We could try assembling a similar function that takes the bindings of a constraint and produces the same bindings with the KVars substituted.

I have an implementation of the KVars substitution in ff7857aea. This is not enough though, because PLE needs to be extended to handle expressions with existentials. Here's the design I'm considering:

1. When PLE first picks a set of bindings in which to look for unfoldings, rename all the existential bindings so they use fresh symbols.
2. When analyzing a candidate expression in search for unfoldings, have PLE remember the existential bindings that it has encountered when traversing the expression in some form of a stack.
3. When PLE finds a new unfolding to make, record together with the new equality, the current stack of existential bindings.
4. When putting the new equalities in the SMT solver environment, use the stacks of existential bindings attached to them, to identify the original existential expressions. Say the original expression was of the form `Ex (p /\ Ey q)`, the new assertion for the predicate environment should be of the form
```
Ex (p /\ new_equalities_under_x /\
    Ey (q /\ new_equalities_under_x_and_y)
   )
```
5. When putting the new equalities back into existing constraints, they should be inserted in constraint environments with the same form as in (4).

In (4) and (5), predicates `p` and `q` might need to be sent multiple times to the SMT solver, we might try simplifying the existential bodies to reduce their size if it turns out to be a problem.

The stack in the design could be avoided. But the removal would cause `p` and `q` to be inserted multiple times in constraints, in cases where they could be inserted only once by properly grouping the equalities they induced.

ranjitjhala:
I suspect this non substitution was to avoid the circular dependency between the two: plugging in KVAR solutions can change (?) result of PLE which can then change result of KVAR solution ...

facundominguez:
[...]

PR #792, opened 2025-11-07
Apply kvar solutions to LHSs before PLE

facundominguez:
This addresses the bug reported in https://github.com/ucsd-progsys/liquidhaskell/issues/2582

As discussed in #780, it is not the best we can do, because we are not looking for unfoldings in kvar solutions, but it is still an improvement.

### DIFF
--- a/src/Language/Fixpoint/Solver/Solution.hs
+++ b/src/Language/Fixpoint/Solver/Solution.hs
@@ -296,19 +296,17 @@ apply g s bs      =
   -- Clear the "known" bindings for applyKVars, since it depends on
   -- using the fully expanded representation of the predicates to bind their
   -- variables with quantifiers.
-  do (ps,  ks) <- envConcKVars g s bs
+  do xrs <- traverse (lookupBindEnvExt g s) (F.elemsIBindEnv bs)
+     let (ps,  ks) = envConcKVars xrs
      (pks, kI) <- applyKVars g {ceBindingsInSmt = F.emptyIBindEnv} s ks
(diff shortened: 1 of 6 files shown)


src/Language/Fixpoint/Solver/Solution.hs
```diff
@@ -296,19 +300,32 @@ apply g s bs      =
...
+--
+applyInSortedReft
+  :: CombinedEnv ann
+  -> Sol.Sol Sol.QBind
+  -> (F.Symbol, F.SortedReft)
+  -> ElabM (F.Symbol, F.SortedReft)
+applyInSortedReft g s xsr@(x, sr) =
+  do let (ps,  ks) = envConcKVars [xsr]
+     (pks, _) <- applyKVars g {ceBindingsInSmt = F.emptyIBindEnv} s ks
+     pure (x, sr { F.sr_reft = F.Reft (x, F.conj (So.unElab pks : ps)) })
```
facundominguez, line 321: I'm calling `unElab`. Maybe I should try instead to stop `applyKVars` from elaborating the term. (?)

facundominguez:

src/Language/Fixpoint/Solver/Solution.hs
facundominguez: For now, this is the smallest that I can do.

It is suspicious that the FUSION implementation is so entangled with elaboration and coercion of set/bag/maps. I'd like to explore separating those.

ranjitjhala: Yes it would be amazing to disentangle those!

PR #795, opened 2025-11-11
Remove and abstract elaboration and sort conversion bits from FUSION

facundominguez:
The goal of this PR is to separate some concerns about elaboration and sort conversions from the FUSION algorithm.

The core changes are:
* QBinds are now elaborated when initializing the KVar solutions. Before it was done every time we wanted to send a predicate to the SMT solver.
* Conversions of sets/maps/bags sorts are now done when elaborating the SInfo before solving. Before, they were done every time we wanted to send a lhs to the SMT solver.
* Elaboration of equalities under existentials is now a function parameter that can be the identity when no elaboration is needed. Before, elaboration was done always, and PLE called unElab.


facundominguez:

cc @clayrat who might be able to check that the changes to sort conversions look reasonable.

src/Language/Fixpoint/SortCheck.hs
```diff
@@ -157,7 +159,7 @@ instance (Elaborate a) => (Elaborate (Maybe a)) where
   elaborate ep t = elaborate ep <$> t
 
 instance Elaborate Sort where
-  elaborate _ = go
+  elaborate ep =  coerceSort (epFlags ep) . go
```
facundominguez, line 162: This has no effect on the tests, but given how elaboration on expressions and SInfo are converting sorts, I think this is the less surprising definition.

facundominguez:
Some small performance differences. Seems to be fine on average.

top
bot

src/Language/Fixpoint/Solver/Solution.hs
```diff
@@ -537,22 +528,14 @@ substSort sEnv sym = fromMaybe (err sym) $ F.lookupSEnv sym sEnv
...
-mkSubst sp env x tx ey ty
-  | tx == ty    = pure $ F.EEq ex ey
-  | otherwise   = do ex' <- elabToInt sp env ex tx
-                     ey' <- elabToInt sp env ey ty
-                     pure $ {- F.tracepp _msg $ -} F.EEq ex' ey'
+mkSubst
+  :: (F.Expr -> F.Expr) -> F.SymEnv -> F.Symbol -> F.Sort -> F.Expr -> F.Sort -> F.Expr
+mkSubst elab env x tx ey ty
+  | tx == ty    = F.EEq ex ey
+  | otherwise   = F.EEq (elab $ So.toInt env ex tx) (elab $ So.toInt env ey ty)
```
facundominguez: Removing `So.toInt` here has no effect on tests either. And this is the only place that uses the `elab` parameter in the whole algorithm.

Checking https://github.com/ucsd-progsys/liquidhaskell/issues/1091 as the comment above hints, seems to imply that ex and ey shouldn't really have different sorts.

Does it look safe to redefine `mkSubst x ey = F.EEq (F.expr x) ey`?

In other words, can we assume that `x` and `ey` have matching sorts?

ranjitjhala: I _think_ so? There were so many kludges put into this code to defensively deal with strange constraints that I can't remember why it was put in the first place :-(

facundominguez: Alright, I removed the elaboration entirely. I also had to remove `tests/pos/bool04.fq` which was building a kvar with ill-typed arguments.

PR #799, opened 2025-11-18
Shrink kvar solutions

facundominguez:
KVar solutions are huge most of the time, and they still show a lot of duplication.

This PR is another attempt to eliminate duplication. The strategy is to transform the pattern `exists x y. x == C && P y` to `exists y. (P y)[x:=C]` repeatedly. This is done both when printing kvar solutions in `.fqout` files and in `Solver.Solution.cubePred`.

With this PR the `fqout` file of `tests/benchmarks/esop2013-submission/Base.hs` goes from 152 klocs to 43 klocs.

The fqout file of the test for #780  shows precisely that `Test.gt0xy $VV##1805##k_ i##aS7` is in the KVar solution without surrounding noise.
```
$k_##1822[$VV##1821##k_ : int,
          $i##aS7##k_ : int,
          $lq_anf$##7205759403792798198##d1ca##k_ : (GHC.Internal.Base.Monad (Test.State int)),
          $lq_anf$##7205759403792798199##d1cb##k_ : (Test.State int Tuple0),
          $lq_anf$##7205759403792798200##d1cc##k_ : (Test.State int int),
          $lq_tmp$x##1823##k_ : int,
          $lq_tmp$x##1824##k_ : int] := exists [w : int]
                                          . Test.gt0xy w i##aS7
                                            && exists [w : int]
                                                 . Test.gt0xy w i##aS7
```

There is some more duplication to eliminate still, because kvar solutions introduce multiple copies of binding predicates. We can eliminate those too, but I'm leaving that to discuss in a separate PR.


facundominguez:
All tests pass in Liquid Haskell. Verification time does not worsen.

src/Language/Fixpoint/Solver/Solution.hs
```diff
@@ -365,16 +353,28 @@ nonCutsResult be s = M.mapWithKey (mkNonCutsExpr g s) $ Sol.sHyp s
...
+dropUnsortedExprs g (F.Su m) = F.Su $
+    M.filter
+      (\e -> isJust $ do
+         t <- So.checkSortExpr sp env e
+         guard (not (isClass t))
+      )
+      m
+  where
+    sp  = ceSpan g
+    env = combinedSEnv g
```
facundominguez, line 379: This was the most surprising bit during the implementation. I do not understand how `liquid-fixpoint` can allow for these unsorted expressions to reach here. `substElim` is also discarding those.

ranjitjhala: Hmm I'm also puzzled. One possibility is polymorphic functions or substitutions, so something like `e` on its own is a well-sorted expression but somehow we end up doing a substitution on it `[x := y]` that does _not_ preserve sorted-ness.

One approach would be to comment out the above code, and see what breaks?

facundominguez: If commenting the code, only one test breaks:
```
cabal exec ghc -- -fplugin=LiquidHaskell -fforce-recomp tests/benchmarks/esop2013-submission/Base.hs
```
The error is not too helpful.
```
**** LIQUID: ERROR *************************************************************
<no location info>: error:
    HasCallStack backtrace:
  collectBacktraces, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:169:13 in ghc-internal:GHC.Internal.Exception
  toExceptionWithBacktrace, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:89:42 in ghc-internal:GHC.Internal.Exception
  throw, called at src/Language/Fixpoint/Types/Errors.hs:151:7 in liquid-fixpoint-0.9.6.3.3-inplace:Language.Fixpoint.Types.Errors
  die, called at src/Language/Fixpoint/Smt/Interface.hs:486:19 in liquid-fixpoint-0.9.6.3.3-inplace:Language.Fixpoint.Smt.Interface

:1:1-1:1: Error
  crash: SMTLIB2 respSat = Error "line 5437 column 4887: unknown constant VV$35$$35$87094"

```

facundominguez: When adding calls to `trace`, `[lq_karg$VV$35$$35$87094##k_##87095:=VV$35$$35$87094]` shows in the substitutions of a kvar, and apparently the bind is not in scope:
```
bind 20663 VV##87094 : {VV##87094 : (Tuple2 k##a478 a##a479) | [$k_##87095]}
```

facundominguez:
I'm merging this and will continue to investigate origin of the out-of-scope binding.

PR #801, opened 2025-11-20
Add a function to save SInfo queries.

facundominguez:
This is intended for debugging, and it is convenient to call as
```
when (save cfg) $ do
  saveSInfo cfg ".sinfo" si
```
This will write a file like `.liquid/Test.hs.sinfo.fq`.

### DIFF
--- a/src/Language/Fixpoint/Types/Constraints.hs
+++ b/src/Language/Fixpoint/Types/Constraints.hs
@@ -932,7 +932,7 @@ saveTextQuery cfg fi = do
 --
 -- Usage example:
 --
--- > when (save cfg) $ do
+-- > when (save cfg) $
 -- >   saveSInfo cfg ".sinfo" si
 --
 -- This will write a file like `.liquid/Test.hs.sinfo.fq` when the
--- a/src/Language/Fixpoint/Types/Constraints.hs
+++ b/src/Language/Fixpoint/Types/Constraints.hs
@@ -925,8 +925,20 @@ saveTextQuery cfg fi = do
   ensurePath fq
   T.writeFile fq $ T.pack $ render (toFixpoint cfg fi)
 
--- Used for debugging to inspect intermediate SInfo files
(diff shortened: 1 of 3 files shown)


src/Language/Fixpoint/Types/Constraints.hs
```diff
@@ -924,6 +925,14 @@ saveTextQuery cfg fi = do
   ensurePath fq
   T.writeFile fq $ T.pack $ render (toFixpoint cfg fi)
 
+-- Used for debugging to inspect intermediate SInfo files
+saveSInfo :: Fixpoint a => C.Config -> FilePath -> SInfo a -> IO ()
```
facundominguez:
```suggestion
-- | Used for debugging to inspect intermediate 'SInfo' files.
--
-- Takes a suffix to put in the name of the written file, whose name
-- is still derived from the input file name in `cfg`.
--
-- Usage example:
--
-- > when (save cfg) $ do
-- >   saveSInfo cfg ".sinfo" si
--
-- This will write a file like `.liquid/Test.hs.sinfo.fq` when the
-- `--save` flag is used.
--
saveSInfo :: Fixpoint a => C.Config -> String -> SInfo a -> IO ()
```

src/Language/Fixpoint/Types/Constraints.hs
```diff
@@ -924,6 +925,26 @@ saveTextQuery cfg fi = do
...
   T.writeFile fq $ T.pack $ render (toFixpoint cfg fi)
 
+-- | Used for debugging to inspect intermediate 'SInfo' files.
+--
+-- Takes a suffix to put in the name of the written file, whose name
+-- is still derived from the input file name in `cfg`.
+--
+-- Usage example:
+--
+-- > when (save cfg) $ do
```
facundominguez:
```suggestion
-- > when (save cfg) $
```

Issue #800, opened 2025-11-20
Uniquifying kvars introduces out-of-scope names


facundominguez:
`wfcUniqify` is introducing names out of scope at this point:
https://github.com/ucsd-progsys/liquid-fixpoint/blob/87b09006a46cca9340e5ff2e9af28d3fa60d8509/src/Language/Fixpoint/Solver/UniqifyKVars.hs#L66

`sym` in that snippet is taken from the environment of the wf constraint of the kvar, but it is not guaranteed to be in the environment of the subtyping constraint where the substitution resides.

This code corresponds to the transformation of a substitution `$k_42[v:=v]` into `$k_42[lq_karg$x$k_42:=x][lq_karg$v$k_42:=v]`, where `x` is listed in the environment of the wf constraint. Note that `x` was initially missing from the domain of the substitution, and after the transformation it appears in its range. 

This behavior doesn't cause a failure because the solver is discarding the substitutions that refer to out-of-scope names. But the question arises: why should substitutions add out of scope names? Which in turn might need revisiting: Why should substitution add to their domain the variables in wf environments?

At the moment, this is more of a request for documentation than a report of something broken.

ranjitjhala:
@facundominguez where does the `x` come from here? Was it in the domain of the WFC for `$k_42`? Where does this constraint come from? If `x` is in the domain of the WfC then really it is a "parameter" of the horn variable `$k_42` so I think it is ill-formed to have just `$k_42[v:=v]` as we don't know what parameter `x` is in that usage site. If we intend the parameter to remain "un-substituted" then the generate substitution should be
`$k_42[lq_karg$x$k_42:=lq_karg$x$k_42][lq_karg$v$k_42:=v]`.

Which `.fq` test are you seeing this behavior on? Thanks for digging into this!

facundominguez:
> @facundominguez where does the x come from here? 

That's correct. I updated the description to make that explicit.

> Which .fq test are you seeing this behavior on?

This is the tests in Liquid Haskell:
```
cabal exec ghc -- -fplugin=LiquidHaskell -fforce-recomp tests/benchmarks/esop2013-submission/Base.hs
```
The role of `x` is taken by `VV##87094` and the kvar is `k_##87090`.

facundominguez:
> I think it is ill-formed to have just $k_42[v:=v] as we don't know what parameter x is in that usage site.

[...]

PR #802, opened 2025-11-20
Describe preconditions on kvar substitutions

facundominguez:
These preconditions are discussed in #800


ranjitjhala (approved):
Super -- this is perfect, thank you!!!

PR #803, opened 2025-11-21
Simplify the solver further

facundominguez:
This PR eliminates `substElim` replacing it with more obvious code.

There is an attempt to produce smaller kvar solutions as well, which I reverted, because the code was not too obvious, didn't win any speedup, and the user would still not see smaller kvars since the simplification routine for `.fqout` files obtains similar results.

But I'm keeping the attempt in the git history for the record.


facundominguez:
All tests pass in Liquid Haskell, verification performance doesn't change.

PR #805, opened 2025-11-24
Remove distinction of predicates and expressions in the parser

### LINKED ISSUE #804: Remove distinction between predicates and expressions in the parser
I'm trying to simplify the parser by removing the distinction between predicates and other expressions. This does make the grammar easier to approach, but I'm finding no way to keep the .fq format unmodified.

The problem manifest at the moment with `qualif` declarations, but I guess `match` declaration will require a similar consideration.

```
qualif IsEmp(v:GHC.Types.Bool, xs: [a]) : <some_expr>

bind 0 n : ...
```

The parser takes the body of `IsEmp` as `<some_expr> bind 0 n`, whereas before, such applications would not be allowed in the `qualif` body.

-------
Possible solutions:

##### make `bind` a reserved word

This fixes the confusion but stops Liquid Haskell from using the identifier in expressions. Other similar words need to be made reserved: qualif, constant, constraint, match, cut, pack, and expand. All of them would be excluded from expressions.

##### put body of `qualif` between delimiters

```
qualif IsEmp(v:GHC.Types.Bool, xs: [a]) : (<some_expr>)
```

or braces

```
qualif IsEmp(v:GHC.Types.Bool, xs: [a]) { <some_expr> }
```

##### terminate body of `qualif` with ";"

```
qualif IsEmp(v:GHC.Types.Bool, xs: [a]) : <some_expr> ;
```

What do you think?

facundominguez:
The parser used to have two forms of expressions: predicates, which must produce a boolean, and the rest of the expressions.

Type checking considerations were handled at parse time, which made the grammar more complicated. This motivated the current PR, which gives up enforcing these typing considerations.

The simplification in turn required introducing delimiters for `qualif` and `match` statements, which required updating the tests, which makes for most of the PR.

Fixes #804


facundominguez:
Tests pass in Liquid Haskell (with modifications to `qualif` annotations).

PR #810, opened 2025-11-28
Always expose solutions of non cut kvars to callers of solve

facundominguez:
The non-cut kvar solutions were hidden if --save was not given (#770).

This PR also generalizes a bit `kvarDomain` so it can be used with FInfo queries.

These changes are needed so kvar solutions can be shown in error messages in Liquid Haskell.

### DIFF
--- a/src/Language/Fixpoint/Parse.hs
+++ b/src/Language/Fixpoint/Parse.hs
@@ -1424,28 +1424,6 @@ crashP pp = do
   msg <- takeWhileP Nothing (const True) -- consume the rest of the input
   return $ Crash [(i, Nothing)] msg
 
-predSolP :: Parser Expr
-predSolP = parens (exprP  <* (comma >> iQualP))
-
-iQualP :: Parser [Symbol]
-iQualP = upperIdP >> parens (sepBy symbolP comma)
-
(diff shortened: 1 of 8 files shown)


facundominguez:
This has an effect on performance as solutions of non-cut kvars are simplified in `simplifyKVars`, and this is slow at least for `tests/neg/StateConstraints0.hs` which still produces very large solutions (40s vs 10s without simplification).

I'm considering how deal with this.

facundominguez:
5 benchmarks worsen a bit in Liquid Haskell when saving fqout files, or when showing solutions of non-cut kvars in error messages (an upcoming PR). I think the overhead is worth a clearer output, and it could be made payable only if verification fails and errors need to be rendered.

Also, this is only noticeable in examples that produce VERY large kvar solutions.

* `tests/neg/StateConstraints0.hs` 7s +8s
* `tests/benchmarks/icfp15/pos/IfM.hs` 2s +5s
* `tests/benchmarks/esop2013-submission/Base.hs` 30s + 3s
* `tests/benchmarks/bytestring-0.9.2.1/Data/ByteString/LazyZip.hs` 2s + 3s
* `tests/pos/StateConstraints.hs` 2s +2s

PR #811, opened 2025-12-03
Allow PLE to unfold in kvar solutions

### LINKED ISSUE #780: Substitute KVars in constraints for PLE

facundominguez:
Closes #780

An overview of the change is in Note [Existential quantification when unfolding] in PLE.hs.

### DIFF
--- a/src/Language/Fixpoint/Smt/Interface.hs
+++ b/src/Language/Fixpoint/Smt/Interface.hs
@@ -106,6 +106,7 @@ import qualified SMTLIB.Backends
 import qualified SMTLIB.Backends.Process as Process
 import qualified Language.Fixpoint.Conditional.Z3 as Conditional.Z3
 import Control.Concurrent.Async (async)
+import GHC.Stack (HasCallStack)
 
 {-
 runFile f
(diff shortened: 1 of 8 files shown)


facundominguez:
All tests pass in Liquid Haskell. Performance doesn't worsen.

ranjitjhala:
Looks great, @facundominguez lets merge!!! this and #812

facundominguez:
Thanks for having a look!

PR #812, opened 2025-12-04
Improvements to creation of apply tags

facundominguez:
This PR contains
* edits to the documentation
* fixes to sendConcreteBindingsToSMT that allow to remove resetAppls
* a small refactoring of funcSortVars


facundominguez:
cc @clayrat

clayrat:
Thanks for picking this up! I suspected that the absence of `pop` in `Solver.PLE.withCtx` was somehow related, but didn't have the energy to investigate it properly.

facundominguez:
I removed the pop for consistency. But the bit that actually required a reset was in `sendConcreteBindingsToSMT`. When returning from the bracket there, the effect of the pop done by the bracket was overwritten by the later call to `put`.

PR #814, opened 2025-12-11
Improve debugging of smt solver queries

facundominguez:
* The .smt2 logging output is annotated with comments that relate it to the code
* simplifyKVar now can be used for SMT queries (it used to call unElab). This is useful for debugging.
* The parser doesn't do simplification before returning expressions. I think I introduced such simplifications accidentally.
* The logging output of PLE is slightly simpler now, as some assertions have less existentials.


ranjitjhala:
This is TERRIFIC -- I was just thinking the other day how nice it would be to have something like it, thanks!!!

PR #815, opened 2025-12-17
Revert "Remove existentials from PLE assertions"

facundominguez:
This reverts commit db97805e2324d602e658596534cf73d131348556.

The reverted changes were making z3 unresponsive in the prover-foundations testsuite (Lists.hs test) of Liquid Haskell.

### DIFF
--- a/src/Language/Fixpoint/Solver/PLE.hs
+++ b/src/Language/Fixpoint/Solver/PLE.hs
@@ -534,21 +534,15 @@ updCtx InstEnv{..} ieSMT ictx delta cidMb mCTrie =
            , icFreshExistentialCounter = existentialCounter
            , icInitialLHSs = M.unionWith S.union candsPerExScopeNoRHS (icInitialLHSs ictx)
            }
-    , ebs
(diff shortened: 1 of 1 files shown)


facundominguez:
z3 is sensitive to how existential quantifications are sprinkled on the assertions.

In the failing test, z3 is happy with `assert (exists ((x Int)) (P x))` but takes ages with
```
declare-fun x () Int
assert (P x)`
```
The bug only shows in large examples so far.

Enabling simplifications to try to overcome the slowness in z3 is buggy as well, causing some queries in other tests to change the response. It is not looking good.

facundominguez:
Hm, actually, I don't think reverting these changes is right. It is worth following up with a bug report to z3. I created https://github.com/ucsd-progsys/liquidhaskell/issues/2599 to track that.

ranjitjhala:
cvc works well here?

- Ranjit.

facundominguez:
That's a good question. cvc5 is not affected by this change. With or without it, it takes 1 minute to verify `Lists.hs`.

PR #816, opened 2025-12-19
Generalize horn syntax to accept expressions for kvars arguments

nilehmann:


ranjitjhala:
The offending test is `arbitrary-kvar-arg.smt2` will take a look as soon as i'm done with the other thing

nilehmann:
@ranjitjhala I did further investigation. 

In [`arbitrary-kvar-arg.smt2`](https://github.com/ucsd-progsys/liquid-fixpoint/pull/816/changes#diff-e3b7233d7111d2161e52920a93c5f3bfc76f1e1b4e3a20ea9babbb01220a0039), the expressions in the kvars' substitution are being dropped here 

https://github.com/ucsd-progsys/liquid-fixpoint/blob/fbe1fcd277c02290be608e35c5081eceaf635c95/src/Language/Fixpoint/Solver/Solution.hs#L378-L388

Removing that filter (i.e., changing the predicate to `True`) fixes the test.

But [`arbitrary-kvar-arg-cyclic.smt2`](https://github.com/ucsd-progsys/liquid-fixpoint/pull/816/changes#diff-8e350c8f9724b9f6181fa968a31ac3b54615a376134a772163792543811111e0) fails for another reason. This test is different because it involves a cycle, so the expressions are being lost somewhere else.

ranjitjhala:
subsumed by #818

PR #820, opened 2026-01-08
Remove old horn parser

nilehmann:
As far as I know, the horn syntax is only used by Flux, and the old parser, which mixed sexp and infix, was kept around for backward compatibility with Flux when the full sexp parser was implemented in https://github.com/ucsd-progsys/liquid-fixpoint/pull/683. 

Flux has been using the new syntax for quite a while, and I don't think it makes sense to maintain two parsers. This PR removes the old parser and the old tests.


facundominguez:
LGTM!

PR #825, opened 2026-03-11
Don't require a space after the comment start

facundominguez:
The space after the comment start was introduced in #445

The errors when a space was missing after // were rather puzzling. e.g.

{-@
//
@-}

produced

Cannot parse specification: incorrect indentation (got 1, should be greater than 1)


facundominguez:
@ranjitjhala, do you remember why the space was needed?

ranjitjhala:
Sadly can't remember but if it involves the parser, I think your intuition is way better than mine so let's merge!

PR #826, opened 2026-03-11
Relax name capture check in substitution of expressions

facundominguez:
Substitution of expressions in https://github.com/ucsd-progsys/liquidhaskell/issues/2625 was failing in a name-capture check. However, the implementation could proceed with the substitution by relaxing the check a bit. This is what the first commit of the PR accomplishes.

The second commit avoids the name-capture check in the particular path that leads to the failure, by using the rapier-style substitution instead.


facundominguez:
The rapier style substitution could introduce some overhead. Before merging we should benchmark the tests to see how well it does.

facundominguez:
There are no significant performance changes in either direction.

top
bot

PR #827, opened 2026-03-12
Add `cut` to horn-parser

ranjitjhala:
So we can mark specific variables as `cut` variables. @vrindisbacher


vrindisbacher:
Testing!

PR #828, opened 2026-03-14
Directly encode String Literals in SMTLIB

ranjitjhala:
The thing where we used a symbol was a bit flaky as sometimes the sort was not declared...

(Claude summary of the bug)

  1. "hello" is parsed as `ESym (SL "hello")` → encoded as symbol `lit$hello` (safe-text: `lit$36$hello`)
  2. `addLiterals` uses `symConsts si` which for `SimpC` only traverses `crhs` — missing `"hello"` when it appears only in LHS binding positions
  3. So `lit$36$hello` never gets a `(declare-fun lit$36$hello () String)` declaration
  4. But it is referenced in the `lhsPred` formula sent to Z3, which crashes with an unknown constant error...
  
A proper cleanup here would be to entirely eliminate the `seLits` machinery perhaps, but I will leave that for another day!

### DIFF
--- a/src/Language/Fixpoint/Smt/Serialize.hs
+++ b/src/Language/Fixpoint/Smt/Serialize.hs
@@ -25,6 +25,8 @@ import qualified Language.Fixpoint.Smt.Theories as Thy
 -- import           Data.Text.Format
 import           Language.Fixpoint.Misc (sortNub, errorstar)
 import           Language.Fixpoint.Utils.Builder as Builder
+import qualified Data.Text as T
+import Data.Text (Text)
 -- import Debug.Trace (trace)
 
 instance SMTLIB2 (Symbol, Sort) where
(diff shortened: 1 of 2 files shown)


src/Language/Fixpoint/Smt/Serialize.hs
```diff
@@ -125,7 +125,7 @@ instance SMTLIB2 LocSymbol where
   smt2 = smt2 . val
 
 instance SMTLIB2 SymConst where
-  smt2 = smt2 . symbol
+  smt2 (SL t) = pure $ quotes $ fromText t  -- emit "hello" not lit$36$hello
```
facundominguez: 👋 This is putting quotes around the literal, but otherwise not doing any attempt to escape characters in the string. Is this what you want?

ranjitjhala: Hmm good point - I was assuming that the SMT solver would just handle the rest; there are no fancy tags or any such. Do we need to escape it somehow? Or do you anticipate some other problem?

facundominguez: > Do we need to escape it somehow? 

I assume SMTLIB must have some escaping rules. Perhaps they are implemented already somewhere in LF.

If it is not escaped, it could be used for code injection attacks, I suppose.

ranjitjhala:
Per the SMTLIB notes I think we need to escape the quote itself

https://smt-lib.org/theories-UnicodeStrings.shtml

(Hmm I wonder what the story is with Unicode…)

facundominguez:
> Per the SMTLIB notes I think we need to escape the quote itself

:+1: Appendix B of the reference says to escape double quotes with double double quotes. (e.g. `a"b` as `a""b`).

PR #831, opened 2026-03-30
chore: add --savebfq flag to collect .bfq files on verification failure

kaf-lamed-beyt:
added a new flag `--savebfq` lag that saves the `.bfq` binary query file only when verification fails (i.e. when the result is `Unsafe`), as described in option 1 of issue [#2020](https://github.com/ucsd-progsys/liquidhaskell/issues/2020) in liquidhaskell.

this makes it possible to collect .bfq files in CI for flaky failures without saving them on every test run.

### DIFF
--- a/src/Language/Fixpoint/Types/Config.hs
+++ b/src/Language/Fixpoint/Types/Config.hs
@@ -249,7 +249,7 @@ defConfig = Config {
   , elimStats                = False   &= help "(alpha) Print eliminate stats"
   , solverStats              = False   &= help "Print solver stats"
   , save                     = False   &= help "Save Query as .fq and .bfq files"
-  , saveBfqOnError           = False   &= help "Save Query as .bfq file only when verification fails" 
+  , saveBfqOnError           = False   &= help "Save Query as .bfq file only when verification fails"
                                        &= name "save-bfq-on-error"
                                        &= explicit
   , metadata                 = False   &= help "Print meta-data associated with constraints"
--- a/src/Language/Fixpoint/Types/Config.hs
+++ b/src/Language/Fixpoint/Types/Config.hs
@@ -250,7 +250,7 @@ defConfig = Config {
   , solverStats              = False   &= help "Print solver stats"
   , save                     = False   &= help "Save Query as .fq and .bfq files"
   , saveBfqOnError           = False   &= help "Save Query as .bfq file only when verification fails" 
-                                       &= name "--save-bfq-on-error"
+                                       &= name "save-bfq-on-error"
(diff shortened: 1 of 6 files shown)


src/Language/Fixpoint/Types/Config.hs
```diff
@@ -248,6 +249,7 @@ defConfig = Config {
   , elimStats                = False   &= help "(alpha) Print eliminate stats"
   , solverStats              = False   &= help "Print solver stats"
   , save                     = False   &= help "Save Query as .fq and .bfq files"
+  , saveBfq                  = False   &= help "Save Query as .bfq file only when verification fails"
```
facundominguez: How about naming this
```haskell
saveBfqOnError = False
 &= help ...
 &= name "--save-bfq-on-error"
 &= explicit
```
?

facundominguez (commented):
Thanks @kaf-lamed-beyt! I left one comment.

In addition, would you like to try enabling this flag in CI and collecting the bfq files in artifacts?

https://docs.github.com/en/actions/concepts/workflows-and-actions/workflow-artifacts

kaf-lamed-beyt:
okay, let me do that now, @facundominguez.

src/Language/Fixpoint/Types/Config.hs
kaf-lamed-beyt: ah, i see. so the `name` explicitly helps us specify the flag. if it was omitted like i had it before, what would've the flag name been, @facundominguez, `--savebfq`?

facundominguez:
I found a couple of important bugs in the test harness of liquid-fixpoint.

The one that hold me for a while was that several tests used the same log file when run in parallel, and this caused failures in my system. I contributed a fix to that in dfc313c.

There are other issues with the tests, that I hope to address in a separate PR. The new flag looks good :+1: I'm going to merge it as soon as CI is green.

There are other pieces necessary for https://github.com/ucsd-progsys/liquidhaskell/issues/2020, but it is ok to do them in separate PRs as well.

kaf-lamed-beyt:
oh okay, @facundominguez. 

> There are other pieces necessary for https://github.com/ucsd-progsys/liquidhaskell/issues/2020, but it is ok to do them in separate PRs as well.

these other pieces... do you mind elaborating? so i can understand and have them in separate PR chunks as you suggested.

facundominguez:
I edited the description of https://github.com/ucsd-progsys/liquidhaskell/issues/2020 with the details.

kaf-lamed-beyt:
great! i'd have a look at it.

PR #834, opened 2026-04-03
Literal Wildcard Patterns in Qualifiers

ranjitjhala:
This PR adds support for literal wildcards, so if you define a qualifier as

```
(qualif Wild2 ((a0 Adt0) (a1# Int) (a2# Str)) (= (Map_select (fld0$1 a0) a1) a2))
```

then the parameters 
- `a1` will get instantiated with any `Int` literals that occur in this constraint, and 
- `a2` with any str literals that occur in the constraint.

cc @vrindisbacher


src/Language/Fixpoint/Solver/Solution.hs
```diff
@@ -61,7 +61,7 @@ initQualifierEnv cfg si
   | scraping  = So.globalEnv cfg si <> instConstants si
   | otherwise = instConstants si
   where
-    scraping = scrape cfg /= No
+    scraping = scrape cfg /= No || allowHOqs cfg
```
ranjitjhala, line 64: this is a bit dicey, will have to check if it ends up causing regressions in LH...

vrindisbacher:
If I understand correctly, in this qualifier: 

```
(qualif Wild2 ((a0 Adt0) (a1# Int) (a2# Str)) (= (Map_select (fld0$1 a0) a1) a2))
```

`a1` and `a2` are wildcards? So fixpoint will essentially churn out all possible `a1` and `a2` that match the `map_select` on the rhs (based on the constraint)?

ranjitjhala:
That's right, so for example, if you run 

```
$ cabal run -- fixpoint tests/horn/pos/wild_lits01.smt2 --save
```

Then you see that

```
Solution:
$k0[$k0##0 : Adt0] := arr_select_m (fld0$1 $k0##0) 1 == (lit "HELLO" Str)
                      && arr_select_m (fld0$1 $k0##0) 2 == (lit "PROFILE" Str)
                      && arr_select_m (fld0$1 $k0##0) 3 == (lit "WORLD" Str)
                      && fld0$0 $k0##0 == 99
```

so its instantiating 
- `a1` with `1`, `2`, `3`, `99` etc., and 
- `a2` with `"HELLO"`, `"PROFILE"`, etc.

ranjitjhala:
@copilot please review this pull request!

PR #837, opened 2026-04-09
Add polymorphic kvar type variable substitution support

facundominguez:
Addresses https://github.com/ucsd-progsys/liquidhaskell/pull/2649

In this solution, liquid-fixpoint substitutes type variables in kvar solutions when they are applied in predicates.

It is the job of liquidhaskell to specify how variables should be substituted.

Each KVar has now a pending substitution of type variables.


facundominguez:
@copilot, please address the hlint warnings in this PR.

PR #839, opened 2026-04-25
Fix polymorphic sort encoding

vrindisbacher:
`fappSmtSort` receives `m` (the number of type parameters in scope) from `sortSmtSort`, but the `Array`, `Set`, and `Bag` cases were recursively calling `sortSmtSort` on sub-sorts without re-wrapping them in `FAbs m`. Each recursive call recomputes `m = sortAbs t` on the unwrapped sub-sort, so a bare `FVar 0` gives `m = -1`, the guard `i < m` fails, and the type variable silently becomes `Int`.

This caused a parametric datatype like:

```
(datatype (Adt 1) ((mk ((fld0 int) (fld1 (Array_t int @(0)))))))
```

to be emitted as `(Array Int Int)` instead of `(Array Int T0)`, producing Z3 sort mismatches when instantiated with non-Int types.

The fix wraps sub-sorts with `FAbs m` before recursing, matching what the `SData` case already did correctly.

### DIFF
--- a/src/Language/Fixpoint/Types/Theories.hs
+++ b/src/Language/Fixpoint/Types/Theories.hs
@@ -336,6 +336,8 @@ fappSmtSort poly m env = go
   where
 -- HKT    go t@(FVar _) ts            = SApp (sortSmtSort poly env <$> (t:ts))
 
+    -- See https://github.com/ucsd-progsys/liquid-fixpoint/pull/839 for why
+    -- @FAbs m@ is re-added.
     go (FTC c) [a]
       | setConName == symbol c   = SSet (sortSmtSort poly env (FAbs m a))
     go (FTC c) [a]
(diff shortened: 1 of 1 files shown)


ranjitjhala:
Awesome- I'll add a regression to catch this and test against LH and then merge.

ranjitjhala:
Amazing -- @cole-k hit exactly the same bug due to something else! (Double thanks @vrindisbacher !)

ranjitjhala:
@facundominguez did you want to look at this? (I have now confirmed that the LH and flux tests pass with this change so I am happy to merge!)

src/Language/Fixpoint/Types/Theories.hs
```diff
@@ -337,13 +337,13 @@ fappSmtSort poly m env = go
 -- HKT    go t@(FVar _) ts            = SApp (sortSmtSort poly env <$> (t:ts))
 
     go (FTC c) [a]
```
facundominguez, line 343:
```suggestion
    -- See https://github.com/ucsd-progsys/liquid-fixpoint/pull/839 for why
    -- @FAbs m@ is re-added.
    go (FTC c) [a]
```

facundominguez (approved):
Looks good to me.

The implementation could be simpler by avoiding the recomputation of `sortAbs t`. It could be done by redefining
```Haskell
sortSmtSort poly = sortSmtSort' poly (sortAbs t)

sortSmtSort' poly m ... =
  <same as current sortSmtSort using m instead of sortAbs>

fappSmtSort = <calls to sortSmtSort' poly m ... >
```

Then introducing `FAbs m` would not be necessary.

PR #842, opened 2026-05-08
Migrate from cmdargs to GetOpt

facundominguez:
This makes simple to customize the help message and to define what flags are available. cmdargs would allow multiple variations of flag names, including prefixes that could be ambiguous (e.g. --save vs --save-bfq-on-error).

### DIFF
--- a/liquid-fixpoint.cabal
+++ b/liquid-fixpoint.cabal
@@ -121,6 +121,7 @@ library
                     Language.Fixpoint.Utils.Progress
                     Language.Fixpoint.Utils.Statistics
                     Language.Fixpoint.Utils.Trie
+                    Language.Fixpoint.Verbosity
                     Text.PrettyPrint.HughesPJ.Compat
   other-modules:    Paths_liquid_fixpoint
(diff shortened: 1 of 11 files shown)


bradjc:
Is it possible this change removed the `--numeric-version` argument? We (Tock) use that to check we have a suitable version to use with our CI:

https://github.com/tock/tock/blob/08c4d1ccc2cf9044ef4af5239d5457122fbb4575/tools/ci/flux-ci-runner/deps.sh#L90

ranjitjhala:
Oops sorry about that let me double check and add that back asap!

facundominguez:
> Is it possible this change removed the --numeric-version argument?

I think so. `cmdargs` provided the flag with the summary [[link]](https://hackage.haskell.org/package/cmdargs-0.10.22/docs/System-Console-CmdArgs-Implicit.html#v:summary), and this escaped my self review.

Hopefully, that's not difficult to add back together with a regression test.

ranjitjhala:
Yes indeed it looks like cmdArgs was automatically generating that numeric-version flag; let me restore it with this new library...

PR #843, opened 2026-05-11
Add explicit numeric version flag

ranjitjhala:


ranjitjhala:
Should fix https://github.com/ucsd-progsys/liquid-fixpoint/pull/842#issuecomment-4424082736

ranjitjhala:
@facundominguez did you want to look at this before I merge?

facundominguez:
LGTM. You might want to add a test for the flag.

The following checks that the parser recognizes the flag, although it has no effect.

```bash
$ cat numeric-version.fq
fixpoint "--numeric-version"
```

The following would also print the numeric version.
```
$ fixpoint --numeric-version
```
but it would need to be coded in the testsuite.

ranjitjhala:
@copilot can you please add a test to the test suite that uses @facundominguez 's strategy to test the `--numeric-version` command line flag?

PR #845, opened 2026-05-22
Change the syms method so it returns a HashSet

facundominguez:
syms was converting to list the result of computing the free symbols as a HashSet.

This PR has syms produce a HashSet instead.

This conversion was sometimes unnecessary. With the upcoming implementation of rapier substitution (https://github.com/ucsd-progsys/liquidhaskell/issues/2676), the interest to use a better representation than lists increases as well (rapier-style substitution requires providing a set with free variables as a parameter).

There is one other commit in this PR to put together the Subable class with the instances in the Substitutions module. For now, this is only to group related code together.

### DIFF
--- a/src/Language/Fixpoint/Solver/Sanitize.hs
+++ b/src/Language/Fixpoint/Solver/Sanitize.hs
@@ -300,10 +300,10 @@ cNoFreeVars fi knownSym c = if S.null fv then Nothing else Just (S.toList fv)
   where
     be   = F.bs fi
     ids  = F.elemsIBindEnv $ F.senv c
-    cDom = [Misc.fst3 $ F.lookupBindEnv i be | i <- ids]
-    cRng = concat [S.toList . F.reftFreeVars . F.sr_reft . Misc.snd3 $ F.lookupBindEnv i be | i <- ids]
-        ++ F.syms (F.crhs c)
-    fv   = (`Misc.nubDiff` cDom) . filter (not . knownSym) $ cRng
+    cDom = S.fromList [Misc.fst3 $ F.lookupBindEnv i be | i <- ids]
+    cRng = S.unions [F.syms . F.sr_reft . Misc.snd3 $ F.lookupBindEnv i be | i <- ids]
+        `S.union` F.syms (F.crhs c)
+    fv   = S.filter (not . knownSym) $ cRng `S.difference` cDom
 
 badCs :: Misc.ListNE (F.SimpC a, [F.Symbol]) -> E.Error
 badCs = E.catErrors . map (E.errFreeVarInConstraint . Bifunctor.first F.subcId)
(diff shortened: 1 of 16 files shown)


facundominguez:
cc @ninioArtillero

src/Language/Fixpoint/Solver/Sanitize.hs
```diff
@@ -300,10 +300,10 @@ cNoFreeVars fi knownSym c = if S.null fv then Nothing else Just (S.toList fv)
...
     be   = F.bs fi
     ids  = F.elemsIBindEnv $ F.senv c
-    cDom = [Misc.fst3 $ F.lookupBindEnv i be | i <- ids]
-    cRng = concat [S.toList . F.reftFreeVars . F.sr_reft . Misc.snd3 $ F.lookupBindEnv i be | i <- ids]
-        ++ F.syms (F.crhs c)
-    fv   = (`Misc.nubDiff` cDom) . filter (not . knownSym) $ cRng
+    cDom = S.fromList [Misc.fst3 $ F.lookupBindEnv i be | i <- ids]
+    cRng = S.unions [F.syms . F.sr_reft . Misc.snd3 $ F.lookupBindEnv i be | i <- ids]
+        `S.union` F.syms (F.crhs c)
+    fv   = S.filter (not . knownSym) $ cRng `S.difference` cDom
```
ninioArtillero, line 306: Flagging that now `cDom` is being filtered as well.

src/Language/Fixpoint/Types/Substitutions.hs
```diff
@@ -363,7 +361,7 @@ meetReft (Reft (v, ra)) (Reft (v', ra'))
 
 instance (Eq v, Hashable v, Refreshable v) => Subable (ReftBV v v) where
   type Variable (ReftBV v v) = v
-  syms (Reft (v, ras))      = v : syms ras
+  syms = reftSymbolsSet
```
ninioArtillero, line 352: `v` (from `Reft (v, ras)`) has been left out of `syms`, effectively replacing `reftFreeVars`; are you sure all calls to `syms` on `ReftBV v v` eventually drop the `v`?

src/Language/Fixpoint/Types/Substitutions.hs
```diff
@@ -170,22 +168,22 @@ instance Subable Symbol where
...
+  syms x                   = S.singleton x
 
 appSubst :: (Eq v, Hashable v) => SubstV v -> v -> ExprBV v v
-appSubst (Su s) x = fromMaybe (EVar x) (M.lookup x s)
+appSubst (Su s) x = M.findWithDefault (EVar x) x s
 
 subSymbol :: (Ord v, Hashable v, Fixpoint v) => Maybe (ExprBV v v) -> v -> v
 subSymbol (Just (EVar y)) _ = y
 subSymbol Nothing         x = x
-subSymbol a               b = errorstar (printf "Cannot substitute symbol %s with expression %s" (showFix b) (showFix a))
```
ninioArtillero, line 159: Why is this error not relevant anymore?

ninioArtillero (commented):
Looks good. Left some comments/questions addressing possible gotchas.

src/Language/Fixpoint/Types/Substitutions.hs
facundominguez: `syms` is supposed to yield the free symbols only, according to the documentation.

In practice each instance returns at least the free symbols, and sometimes more. The tests do not seem to be sensitive to the extra symbols.

A follow up change will return only the free symbols, so I'm not much concerned about this variation.

src/Language/Fixpoint/Types/Substitutions.hs
facundominguez: Good catch. I think it is immaterial. I added another commit to remove the function and the `Subable Symbol` instance which is not used.

src/Language/Fixpoint/Solver/Sanitize.hs
ninioArtillero: Oh, now I see it is equivalent as the filtering ends up being only on `cRng` elements anyway.

PR #849, opened 2026-08-08
Change div/mod definition to match Haskell rounding semantics

tsung-ju:
This is a follow-up to ucsd-progsys/liquidhaskell#2733

I changed the `/` and `mod` operators to generate `HASKELL_DIV` and `HASKELL_MOD` respectively, which are defined as:

```z3
(define-fun HASKELL_DIV ((x Int) (y Int)) Int (ite (> y 0) (div x y) (- (div (- x) y)))
(define-fun HASKELL_MOD ((x Int) (y Int)) Int (ite (> y 0) (mod x y) (- (mod (- x) y)))
```

On the liquid-haskell side, once this is merged, `/` `mod` becomes the correct definitions for `div` and `mod`:
```haskell
define div x y        = (x / y)
define mod x y        = (x mod y)
```
On the other hand, definitions for `quot` and `rem` needs to be updated to e.g.
```haskell
define quot x y       = if (x >= 0) == (y >= 0) then abs x / abs y else -(abs x / abs y)
define rem x y        = if x >= 0 then (abs x) mod (abs y) else -((abs x) mod (abs y))
```

### DIFF
--- a/src/Language/Fixpoint/Smt/Serialize.hs
+++ b/src/Language/Fixpoint/Smt/Serialize.hs
@@ -155,7 +155,7 @@ instance SMTLIB2 Bop where
   smt2 Div    = pure $ symbolBuilder divFuncName
   smt2 RTimes = pure "*"
   smt2 RDiv   = pure "/"
-  smt2 Mod    = pure "mod"
+  smt2 Mod    = pure $ symbolBuilder modFuncName
 
 instance SMTLIB2 Brel where
   smt2 Eq  = pure "="
--- a/src/Language/Fixpoint/Smt/Theories.hs
+++ b/src/Language/Fixpoint/Smt/Theories.hs
@@ -244,16 +244,19 @@ bFun' name ts out = key "declare-fun" (seqs [fromText name, args, out])
 bSort :: Raw -> Builder -> Builder
 bSort name def = key "define-sort" (fromText name <+> "()" <+> def)
 
-
+-- Generate division/modulo function that matches Haskell's rounding behavior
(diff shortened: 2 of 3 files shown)


facundominguez:
Thanks @tsung-ju. I think a change to the meaning of integer division in `liquid-fixpoint` will require some more justification.

It makes sense to adjust the meaning of div and mod in Liquid Haskell so they match Haskell's operations. `liquid-fixpoint` is not always used for Haskell though. Should everybody else accommodate to Haskell nuances?

https://github.com/ucsd-progsys/liquidhaskell/issues/2547 is quite good to address the problem in LH. It looks to me like we could change how `/` and `mod` are exposed in LH so they are not confused with the operations matching Haskell's meaning. For instance, we could rename them to `lfMod` and `lfDiv` :shrug:

ranjitjhala:
To maintain compatibility, we could add a flag `--haskell-div` or similar to `liquid-fixpoint` and use that to gate the interpretation of div/mod? [ Apologies if this has already been discussed! ]

tsung-ju:
Thanks for clarifying! I wasn't aware that `liquid-fixpoint` is also used by other languages. That's very cool!

In that case, I agree that we should consider removing the hard-coded `/` and `mod` from the syntax and instead exposing them under more explicit names. Even in other languages, I think the Euclidean definition is rarely the default and could lead to similar confusion. Perhaps we could borrow Rust's `div_euclid` / `rem_euclid` naming?

Then, in Liquid Haskell, we could define `/` and `mod` as infix operators with Haskell semantics.

Changing the syntax is a bit beyond my capacity at the moment, though. I'll close my PRs for now, but please feel free to recycle any code you find useful.

PR #852, opened 2026-09-02
Support unicode syntax

vishwaksen-1:
To help fix LH Issue - ucsd-progsys/liquidhaskell#2551 'Liquid Haskell should accept UnicodeSyntax'.

Adding the changes to parser from this repo, and those in LH tests in LH repo.

Currently only added aliasing for unicode representations of `::`, `->`, `=>` as I found only those in the project, unlike Haskell which has support for other unicode characters as well.

Please check.

Thank you.

### DIFF
--- a/src/Language/Fixpoint/Parse.hs
+++ b/src/Language/Fixpoint/Parse.hs
@@ -546,6 +546,8 @@ _reservedOpNames =
   , "|-"
   , "::"
   , "."
+  -- unicode aliases
+  , "∷", "⇒", "→"
   ]
 
 {-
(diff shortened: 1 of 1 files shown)


vishwaksen-1:
I'm updating the tests file as a different PR in LH repo.
Tests are all green now.

Thank you. :)

src/Language/Fixpoint/Parse.hs
```diff
@@ -602,12 +604,23 @@ locReserved x =
 --
 reservedOp :: String -> ParserV v ()
 reservedOp x =
-  void $ lexeme (try (string x <* notFollowedBy opLetter))
+  void $ lexeme (try (string x <* notFollowedBy opLetter)
+    <|> try (string (unicodeAlias x) <* notFollowedBy opLetter))
```
facundominguez, line 608: Maybe the alternative should be added only if there is a unicode alias for the operator.

facundominguez (commented):
Thanks @vishwaksen-1.

This is adding some unicode operators to `liquid-fixpoint`. I don't have a strong opinion on whether this should be done here, affecting all users of `liquid-fixpoint`, or only for LiquidHaskell. Do you have any opinion on this regard?

src/Language/Fixpoint/Parse.hs
vishwaksen-1: 😅 I was just relying on `try` to take care of it

facundominguez:
Probably this PR should add a test for the new operators. There are tests for the parser in
`tests/tasty/ParserTests.hs`, and some small fixpoint query file could be added in `tests/pos` that uses the unicode operators.

src/Language/Fixpoint/Parse.hs
```diff
@@ -602,12 +604,23 @@ locReserved x =
...
+  where
+    unicodeAlias "::"     = "∷"  -- U+2237
```
facundominguez, line 610: I noted that the parser has an occurrences of `"::"` in the definition of `dcolon`. Probably `dcolon` should allow the unicode form as well.

vishwaksen-1:
Hi @facundominguez Yes I'll look into them and update you.

src/Language/Fixpoint/Parse.hs
vishwaksen-1: Yes I just realised I totally missed it. 
Also I just found by grepping that - `dcolon` and `arrow` are used for `->` and `::` in `./src/Language/Fixpoint/Misc.hs`. And some instances of `Arrow` (I'm not sure if this refers to `->` I have to check. )

vishwaksen-1:
I tried pushing and it went to a new PR. please don't mind 😅

facundominguez:
Superseded by #853.

PR #853, opened 2026-09-05
Support unicode syntax

vishwaksen-1:
I've added the tests and modified the instance at `dcolon`.
Also checked for other places where the unicode usage of `::, ->, =>` might matter. I think these are all of them.
Please check and give me your feedback.

Thank you.

### DIFF
--- a/src/Language/Fixpoint/Parse.hs
+++ b/src/Language/Fixpoint/Parse.hs
@@ -434,7 +434,7 @@ semi, comma, colon, dcolon, dot :: ParserV v String
 semi   = sym ";"
 comma  = sym ","
 colon  = sym ":" -- Note: not a reserved symbol; use with care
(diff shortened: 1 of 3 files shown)


facundominguez (approved):
LGTM!

vishwaksen-1:
Thank you for the changes. 😅 I feel a little embarrassed that my edits weren't as elegant as they could have been. I'll keep improving and try to get my changes closer to the project's style and standards.

Also, could you please tell me a little about what the new `.fq` test is doing? I'd like to understand it better.

Thank you. :)

facundominguez:
> Also, could you please tell me a little about what the new .fq test is doing?

This test is a copy of another one with the unicode operators inserted. This particular test is enabling flags to reason with functions and checking a single constraint with left hand side `true ⇒ true` and right hand side `f = \y : int → (13 ∷ int)` which is an equality between functions.

The README of liquid-fixpoint gives some clues about the format, but proper documentation is still due.

